PR #162, opened 2024-06-27
feat: introduce standalone mode

emersion:
# Description

This is the first part of #158, cleaned-up and rebased.

This is missing inputs/outputs for `AppComponent`: that work will come as a second follow-up PR once this one is merged.

**The normal behavior of the application remains the same** (`ngsscbuild` has not been modified, only scripts to trigger new `standalone` environment have been added)

# Checklist


src/app/netzgrafik-application/netzgrafik-application.component.ts
```diff
@@ -48,13 +52,24 @@ export class NetzgrafikApplicationComponent {
...
           uiInteractionService.getDefaultViewProperties());
       });
   }
 
+  getSidebarContainerStyle(): string {
+    if (this.disableBackend) {
+      return "top: 53px;";
+    }
+    return "top: 85px;";
+  }
```
emersion, line 71: Hm, I wonder if we should instead add a `disable-backend` CSS class, and use that to guard the CSS rules, instead of this?

src/app/netzgrafik-application/netzgrafik-application.component.ts
```diff
@@ -48,13 +52,24 @@ export class NetzgrafikApplicationComponent {
       .subscribe((params) => {
         uiInteractionService.setEditorMode(EditorMode.NetzgrafikEditing);
         uiInteractionService.showNetzgrafik();
-        versionControlService.load(params.getVariantId(), true);
+        try {
```
emersion, line 55: Is there a reason why this is a `try`/`catch` instead of a `if (disableBackend)`?

src/app/services/auth/auth.service.ts
```diff
@@ -30,27 +30,29 @@ export class AuthService {
...
-    // Redirect the user to the url configured with state above or in a separate login call.
-    this.oauthService.events
-      .pipe(first((e) => e.type === "token_received"))
-      .subscribe(() => {
-        const state = decodeURIComponent(this.oauthService.state || "");
-        if (state && state !== "/") {
-          this.router.navigate([state]);
-        }
-      });
+    if (!environment.disableBackend) {
```
emersion, line 33: This could be simplified with an early return.

src/environments/environment.model.ts
```diff
@@ -3,6 +3,8 @@ import {AuthConfig} from "angular-oauth2-oidc";
 export interface Environment {
   production: boolean;
   label: string;
-  backendUrl: string;
-  authConfig: AuthConfig;
+  backendUrl?: string;
+  authConfig?: AuthConfig;
+  disableBackend: boolean;
```
emersion, line 8: We've been wondering whether it would make more sense to drop the `disableBackend` field, and instead make it so a null `backendUrl` means that the backend is disabled. Personally I think that'd make sense since `backendUrl` is nullable.

src/app/services/auth/auth.service.ts
louisgreiner: agreed and fixed

louisgreiner:
Precisions:

This PR creates a new environment for the application

louisgreiner (approved):

aiAdrian (approved):
This part can be merged

emersion:
Thank you!

Issue #168, opened 2024-07-04
[Bug]: Rounding issues with decimal numbers.


aiAdrian:
### Preflight Checklist

- [X] I have read the [Contributing Guidelines](https://github.com/SchweizerischeBundesbahnen/netzgrafik-editor-frontend/blob/main/CONTRIBUTING.md) for this project.
- [X] I agree to follow the [Code of Conduct](https://github.com/SchweizerischeBundesbahnen/netzgrafik-editor-frontend/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://github.com/orgs/sbb-design-systems/projects/1) for an issue that matches the one I want to file, without success.

### Bug type

Functionality

### Which version are you using

2.5.0

### Operating system

all

### Browser / Browser version

all 

### Input mode

None

### What happened?

When using numbers other than just even numbers, "very specific" digits come into play after the decimal point. Everything is fine with 0.5 minutes, but other decimal values are incorrect and have consequences. This is an unresolved case/pending issue for a long time.  

![image](https://github.com/SchweizerischeBundesbahnen/netzgrafik-editor-frontend/assets/10423646/ee062a0d-daad-4869-acd0-d919728091a4)

### Steps to reproduce the issue

At the moment no clear what happens.

aiAdrian:
The issue is caused by the % operator !!!! 

The remainder (%) operator returns the remainder left over when one operand is divided by a second operand. It always takes the sign of the dividend.

To test: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Remainder

 
`console.log(13 % 5);
// Expected output: 3

console.log(-13 % 5);
// Expected output: -3

console.log(4 % 2);
// Expected output: 0

console.log(-4 % 2);
// Expected output: -0

console.log(-4.33 % 2);
// Expected output: -0.33 but -0.33000000000000007 got !!!

console.log(-4.25 % 2);
// Expected output: -0.25
`

PR #175, opened 2024-07-11
Use epsilon comparison instead of rounding

emersion:
Use the formula `abs(a - b) < epsilon` to check whether two floating-point values are equal.

### DIFF
--- a/src/app/models/trainrunsection.model.ts
+++ b/src/app/models/trainrunsection.model.ts
@@ -15,7 +15,6 @@ import {
 } from "../data-structures/technical.data.structures";
 import {TrainrunsectionValidator} from "../services/util/trainrunsection.validator";
 import {formatDate} from "@angular/common";
-import {MathUtils} from "../utils/math";
 
 export class TrainrunSection {
   private static currentId = 0;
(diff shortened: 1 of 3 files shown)


aiAdrian:
Looks good for the validation side, but does it also resolve the epsilon issue when the user enter any floating point number, for example: 
he like to enter 

128.4 = 2h 8min und 24s => console.log((128.4) % 60); => 8.400000000000006 which is not correct w.r.t. business 8.4 would be fine. but the 8.40000000000000**6** must be avoided.

louisgreiner:
I does look good for validation and visualisation in netzgrafik and perlenkette components, even with weirder figures and changing the number of decimal displayed

emersion:
@aiAdrian, what do you mean exactly? Are you concerned about 8.400000000000006 getting saved into the store/database?

I personally think we should never care about rounding issues in internal code (intermediary values, computations, etc), and we should only care about them right at the end, when we actually display the value to the user (in that case, we want to round according to `getTimeDisplayPrecision()`) or compare it (in that case, compare with an epsilon with the formula in this PR).

Does that make sense?

aiAdrian:
Yes, indeed i care about storing those values. Especially they make no sense. Under the assumption the stored value gets further exported to 3rd party it would good having at least in the database (JSON) "correct" data. This isn't related to the solved validation question. The addressed problem gets solved with #168

PR #206, opened 2024-08-02
refactor: simplify formatDisplayText()

emersion:
Use a loop instead of repeating the replace() calls.


louisgreiner (approved):
nice refactor, thanks !

PR #210, opened 2024-08-05
fix: disable links section in node sidebar

emersion:
This is SBB-specific.

### DIFF
--- a/src/app/view/editor-side-view/editor-node-detail-view/editor-node-detail-view.component.html
+++ b/src/app/view/editor-side-view/editor-node-detail-view/editor-node-detail-view.component.html
@@ -273,7 +273,7 @@
       </tr>
     </table>
   </sbb-expansion-panel>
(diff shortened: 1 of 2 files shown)


louisgreiner (approved):

aiAdrian (approved):

aiAdrian:
I approved this. But we might have to refactor once the disableBackend approach. This changed might be an option which could only be used if the operation point is located in switzerland. => At the moment i don't have any idea how to resolve this issue in a more general manner.

emersion:
That's a good point! We could add a separate knob to enable/disable SBB-specific features if we wanted to. Then from standalone mode one could do `<sbb-root disable-sbb-features="true">` if they want to.

louisgreiner:
I opened a new issue to track this and work on it later

-> https://github.com/SchweizerischeBundesbahnen/netzgrafik-editor-frontend/issues/216

PR #212, opened 2024-08-05
refacto trainrun filter service for label uniqueness

louisgreiner:
# Description

When using (create, delete, focusout) the trainrun label area, the trainrun label service makes unnecessary calls.

This feature:
- test uniqueness of `labels` before calling `setLabels()`
- remove double calls when `create` a label, but also `focusout` at the same time, thus calling twice the functions

Update:  does the same refactor for node filters

# Issues

# Checklist

### DIFF
--- a/src/app/view/dialogs/note-dialog/note-filter-tab/note-filter-tab.component.ts
+++ b/src/app/view/dialogs/note-dialog/note-filter-tab/note-filter-tab.component.ts
@@ -30,9 +30,11 @@ export class NoteFilterTabComponent implements OnInit, OnDestroy {
 
   public note: Note;
   public noteLabels: string[];
+  private initialNoteLabels: string[];
   noteLabelsAutoCompleteOptions: string[] = [];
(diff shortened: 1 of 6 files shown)


louisgreiner:
Done as well for Note and fix another bug in the same component.

Please test and merge if functionnal

src/app/view/dialogs/note-dialog/note-filter-tab/note-filter-tab.component.ts
```diff
@@ -64,7 +71,9 @@ export class NoteFilterTabComponent implements OnInit, OnDestroy {
     this.noteLabels = this.noteLabels.filter(
       (labels) => labels !== valueDelete,
     );
-    this.noteService.setLabels(this.note.getId(), this.noteLabels);
+    this.isLabelBeingEdited = true;
```
emersion, line 74: Hm, do we really need these booleans? Can't we pass a flag via function arguments instead?

How does `onLabelsFocusout` end up being called by `checkAndSetLabels`?

PR #230, opened 2024-08-08
fix: disable environment header in standalone mode

emersion:
According to the [sbb-angular docs][1], the sbb-header-environment component should be disabled in production. Do so in standalone mode.

Without this, a ribbon with a "standalone" text is displayed in the top-left corner.

out

[1]: https://github.com/sbb-design-systems/sbb-angular/blob/main/src/angular/header-lean/header-lean.md#environment


aiAdrian (approved):

emersion:
Thanks for the review!

PR #238, opened 2024-08-12
236 feature request security updates

aiAdrian:
# Description

The Dependabot reports security problems. The packages used should therefore be kept to a minimum and those installed should also be up to date in order to minimise the risk of security problems (virtualities).
 
# Issues

[Dependabot - alerts](https://github.com/SchweizerischeBundesbahnen/netzgrafik-editor-frontend/security/dependabot)
# Checklist


emersion (approved):
LGTM

Note that most of the alerts sound like false alarms: they are about server-side stuff we don't care about since this is the frontend, or devDependencies which shouldn't matter since they're only run by developers.

PR #248, opened 2024-08-19
fix: use change event instead of keyup for node name change

emersion:
Fixes a few issues:

- "keyup" is not triggered when copy-pasting. So if the user right clicks and pastes a new value, the name would not get updated.
- "keyup" triggers each time a character is typed in the input field. This results in a lot of operations on each keypress.
- No need to set this.nodeProperties.nodeBetriebspunktName nor this.nodeProperties.nodeBetriebspunktFullName on keyup: this is already done automatically by [ngModel].

Note that we do loose a bit of functionality with this patch: the diagrams and sidebar headings aren't updated in real-time as the user types anymore. If this is deemed important functionality, we could use `(ngModelChange)` instead.

### DIFF
--- a/src/app/view/editor-side-view/editor-node-detail-view/editor-node-detail-view.component.html
+++ b/src/app/view/editor-side-view/editor-node-detail-view/editor-node-detail-view.component.html
@@ -10,14 +10,14 @@
       <input
         sbbInput
         [ngModel]="nodeProperties.nodeBetriebspunktName"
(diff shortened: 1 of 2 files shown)


louisgreiner:
Great idea :+1:

louisgreiner (approved):

aiAdrian (approved):
Thank you.

PR #250, opened 2024-08-19
244 bug viewport not centering on bounding box when reloadingopening netzgrafik

aiAdrian:
# Description

The viewbox center gets centered when:
- Reloading
- Opening a new project/variant
- Uploading a netzgrafik from JSON

In all other cases, the bounding box of the current netzgrafik has no impact on the centering, which means the view center does not automatically align. It remains as it is, even when the preview of a version is called or when the graphical timetable (Streckengrafik) is opened/closed. When the preview of a version is called and the temporarily loaded netzgrafik is no longer in the viewbox, it will also not be automatically re-centered – even if "discard changes" is called.# Issues

# Checklist

# Test required 
* [ ] Reviewer has to test this bug fix [feature/enhancement must be well tested]


src/app/view/util/svg.mouse.controller.ts
```diff
@@ -168,6 +168,10 @@ export class SVGMouseController {
   }
 
   moveNetzgrafikEditorViewFocalPoint(center: Vec2D) {
+    if (this.viewboxProperties === undefined) {
+      console.log("moveNetzgrafikEditorViewFocalPoint", null);
```
louisgreiner, line 172: Don't forget to remove `console.log` before merging, if it's for test purpose

louisgreiner (approved):
Except the last comment, LGTM and tested :+1:

src/app/view/util/svg.mouse.controller.ts
aiAdrian: Fixed - thank you

Issue #191, opened 2024-07-22
[Feature request]: New mode: integrated-mode (no back-end but UI edition allowed)


louisgreiner:
### Preflight Checklist

- [X] I have read the [Contributing Guidelines](https://github.com/SchweizerischeBundesbahnen/netzgrafik-editor-frontend/blob/main/CONTRIBUTING.md) for this project.
- [X] I agree to follow the [Code of Conduct](https://github.com/SchweizerischeBundesbahnen/netzgrafik-editor-frontend/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://github.com/orgs/sbb-design-systems/projects/1) for an issue that matches the one I want to file, without success.

### Request type

Request for enhancement of a component

### Functionality

Request for a new mode that:
- doesn't need back-end
- enables edition in UI (to trigger the [`Operation` events](https://github.com/SchweizerischeBundesbahnen/netzgrafik-editor-frontend/blob/9d2c6eee74f94912b68ec8b3375dc53e21a5ecdc/src/app/models/operation.model.ts))

The write mode needs a back-end to work properly. The read-only mode is useful for network graphic display with restricted permissions. This mode is used for the "standalone" environment (= integrated in another application) and allows to work without back-end. However the current read-mode is bugged ([see issue](https://github.com/SchweizerischeBundesbahnen/netzgrafik-editor-frontend/issues/190)) and once it is fixed, the "standalone" environment will not allow to write anymore, which will break its use in other applications.

louisgreiner:
Don't seem to need it still, at least postponed

PR #257, opened 2024-08-20
fix: Bug archived read only mode variants are editable even tho not persisted

aiAdrian:
# Description

When the project is archived, the user can no longer modify the netzgrafik. All editing functions should be disabled.

# Testing

This bug fix is a big one. Therefore, we have to deeply test the user interface interactions. There should be no differences from the current version as long as the project is not archived. If a project along with their variants are archived, the display of these variants is read-only. That means the user can no longer make any changes.

* [ ] All manual tests passed
 

# Checklist


aiAdrian:
> I will take a close look to this PR. Have you seen this related issue ? #191 The issue is that NGE integrated version currently runs with the bugged `read-only` mode (so yet not fixed by this PR). By merging this PR, the integrated versions of NGE will be broken. Had you have the time to think about a 3rd mode yet ?

Thank you for your comment. I believe integrating the standalone application will be quite simple. If environment.backendDisabled is set to true, the getVariantIsWritable method in VersionControlService should always return true. Additionally, the environment object could be extended with a flag - readonly: boolean. If readonly is set to true, everything can be handled in VersionControlService to enforce the read-only mode. Possibly, this method should be moved to another service, which requires a software architectural review that has not yet been done in detail. However, this can be easily achieved through refactoring.

louisgreiner:
image

It would be nice to hide this grey bar on bottom on button where the mouse is (hover) ? Disable this bar when a button is disable

aiAdrian:
I have overworked the styling and the behavior (only styling) when:
- readonly
- if Streckengrafik is on
- if Analytics is on
for the menu. It should now be finished/polished. But please re-test. Thanks.

louisgreiner:
Shouldn't archived projects **automatically** archive variants that aren't archived yet ?

image

Here, the variant isn't archived, even tho the project is, but the if we open the variant, the mode is archived

image

The solution would be to automatically archive all the variants of a project that is archived, and tick and disable by default the Show archive button ?

image

src/app/perlenkette/perlenkette-node/perlenkette-node.component.html
```diff
@@ -1,4 +1,6 @@
-<div class="node-container" (click)="disapleSectionView($event)">
+<div class="node-container" (click)="disapleSectionView($event)"
```
SergeCroise, line 1:
```suggestion
<div class="node-container" (click)="disableSectionView($event)"
```

aiAdrian:
> Shouldn't archived projects **automatically** archive variants that aren't archived yet ?
> 
> image
> 
> Here, the variant isn't archived, even tho the project is, but the if we open the variant, the mode is archived
> 
> image
> 
> The solution would be to automatically archive all the variants of a project that is archived, and tick and disable by default the Show archive button ?
> 
> image

I'm not sure if I understood everything correctly. So I would like to briefly explain what the archive concept looks like:

If a project is archived, then all attached (assigned) variants are read-only. (Here I have just identified a problem, i.e., if a project is archived, there is an auth error if one tries to archive a non-archived variant -> read-only protection applies). However, the variants are not shown as archived.

If a project is not archived, variants can be archived, as the project is not read-only. However, this requires that the user has write permissions.

Archiving means deleted – in other words, only archived variants can be deleted. The same applies to projects. It’s a two-step process, so that on one hand, the user can be really sure that they want to delete something. Or one can define in the process that all work should remain saved, then everything that is no longer needed is archived and that’s it.

aiAdrian:
> > Shouldn't archived projects **automatically** archive variants that aren't archived yet ?
> > ...
> > The solution would be to automatically archive all the variants of a project that is archived, and tick and disable by default the Show archive button ?
> 
> I'm not sure if I understood everything correctly. So I would like to briefly explain what the archive concept looks like:
> 
> If a project is archived, then all attached (assigned) variants are read-only. (Here I have just identified a problem, i.e., if a project is archived, there **is an auth error if one tries to archive** a non-archived variant -> read-only protection applies). However, the variants are not shown as archived.
> 
> ...

The above reported ***error fixed*** is fixed : https://github.com/SchweizerischeBundesbahnen/netzgrafik-editor-frontend/pull/257/commits/4b6c5e52dc8680f973f5726eeac355ffd2e96167

src/app/perlenkette/perlenkette-node/perlenkette-node.component.html
aiAdrian: Should be fixed. Thanks

src/app/view/editor-main-view/data-views/trainrunsections.view.spec.ts
```diff
@@ -171,7 +171,8 @@ describe("TrainrunSection-View", () => {
       copyService,
       logService,
       viewportCullSerivce,
```
SergeCroise, line 173: ?

louisgreiner:
Thanks for your explanations in [your previous comment](https://github.com/SchweizerischeBundesbahnen/netzgrafik-editor-frontend/pull/257#issuecomment-2311650809), I better understand the logic : the variants inside an archived project should not be able to be (un)archived afterwards. The user can then see the archived and unarchived variants of an archived project -> working well

louisgreiner (approved):

PR #325, opened 2024-10-24
fix: 320 bug graphical timetable streckengrafik renders only one trainrun segement

aiAdrian:
# Description

This change fixes the problem if a trainrun does not have a continuous path from the start to the destination, i.e. if sections are missing in the path that should continuously connect the nodes from the start to the end. These **holes** are artifacts that can occur during drawing if the train has not been drawn correctly or has not yet been fully drawn. So there can always be a train run that has "holes" along the path, i.e. missing train run sections along the path, e.g.
A - B - C ---- missing --- E - F - G --> also no section between C and E.

The solution to the problem - described above - is proposed in this fix: The trainrun has two parts: [(A B), (A C)] and [(E F)(F G)]. The trainrun is now interpreted as two separate train runs. The network diagram provides no information as to whether the missing part is between C - E, C - D - E, or another possible variant. Thus the best assumption we can make here is that each part of the train run can be interpreted as a single trainrun. 

(In principle, however, this should not occur in a perfect Netzgrafik.)

image
Example Netzgrafik 

image
Graphical timetable 

# Issues fixed

The issue was fixed in the **graphical timetable (Streckengrafik)** within the TrainrunSection collector. All trainruns are now selected using the default iteration, which starts at the top-left end and traverses through the trainrun. Previously, if a trainrun had more than one segment the old version only collected a "random" trainrun part. This has now been fixed.

Additionally, the **cumulativeTimePropagation** for the trainrun had a similar issue, which caused other problems in the graphical timetable. This has also been resolved. Each trainrun segment is now interpreted as a single trainrun and is propagated separately.

The trainrun part issue was as well found in the **pearlsview** -  it's as well resolved.
image

# Checklist

### DIFF
--- a/src/app/streckengrafik/services/sg-1-load-trainrun-item.service.ts
+++ b/src/app/streckengrafik/services/sg-1-load-trainrun-item.service.ts
@@ -394,19 +394,15 @@ export class Sg1LoadTrainrunItemService implements OnDestroy {
     const bothEndNodes =
       this.trainrunService.getBothEndNodesFromTrainrunPart(trainrunSection);
 
-    let startForwardNode = GeneralViewFunctions.getLeftOrTopNode(
+    const startForwardBackwardNode = GeneralViewFunctions.getStartForwardAndBackwardNode(
       bothEndNodes.endNode1,
       bothEndNodes.endNode2,
     );
-    let startBackwardNode =
-      bothEndNodes.endNode1.getId() === startForwardNode.getId()
-        ? bothEndNodes.endNode2
-        : bothEndNodes.endNode1;
 
     if (onlyForward) {
       const tsg = this.trainrunSectionGroup(
         trainrun.getId(),
-        startForwardNode,
+        startForwardBackwardNode.startForwardNode,
         true
       );
       this.forwardTrainrunSectionGroup = tsg.trainrunSectionGroups;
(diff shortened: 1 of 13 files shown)


src/app/streckengrafik/services/sg-1-load-trainrun-item.service.ts
```diff
@@ -624,33 +669,56 @@ export class Sg1LoadTrainrunItemService implements OnDestroy {
...
-            trainrunItem,
-            templateTrainrunItem,
-          );
-          trainrunItems.push(trainrunItem);
+          // get all trainrun section for given trainrun
+          let alltrainrunsections = this.trainrunSectionService
+            .getAllTrainrunSectionsForTrainrun(trainrun.getId());
+
+          // As long not all trainrun section are visited (process) continue
+          // this part of code supports partial cancelatlations, e.g. trainrun runs from
```
SergeCroise, line 703:
```suggestion
          // this part of code supports partial cancellations, e.g., trainrun runs from
```
see
https://www.merriam-webster.com/thesaurus/cancellation
and
https://www.merriam-webster.com/dictionary/e.g.#examples

aiAdrian:
Iterator pattern for iterating through all trainrun parts seperatly

```
simpleTrainrunIterator(trainrun: Trainrun) {
    let alltrainrunsections =
      this.trainrunSectionService
        .getAllTrainrunSectionsForTrainrun(trainrun.getId());

    while (alltrainrunsections.length > 0) {
      // traverse over all trainrun parts
      const trainrunSection = alltrainrunsections[0];
      const bothEndNodes =
        this.trainrunService.getBothEndNodesFromTrainrunPart(trainrunSection);
      const startForwardNode = GeneralViewFunctions.getLeftOrTopNode(
        bothEndNodes.endNode1,
        bothEndNodes.endNode2,
      );

      const startTrainrunSection = startForwardNode.getStartTrainrunSection(
        trainrun.getId(),
      );

      const iterator: TrainrunIterator = this.trainrunService.getIterator(
        startForwardNode,
        startTrainrunSection,
      );

      while (iterator.hasNext()) {
        const currentTrainrunSectionNodePair = iterator.next();
        const trainrunSection = currentTrainrunSectionNodePair.trainrunSection;
        const node = currentTrainrunSectionNodePair.node;

      ...
        user
        defined
        data
        processing
      ...

        // filter all still visited trainrun sections
        alltrainrunsections = alltrainrunsections.filter(ts =>
          ts.getId() !== trainrunSection.getId()
        );
      }


    }
  }
```

PR #349, opened 2024-11-13
Refacto language + add navigator preferences

louisgreiner:
# Description

- Initialize application language with navigator language if no preference (= first connection) in localStorage
- Allow to change language from the application without window reload
- Allow to change language from the outer of the application with `@Input (setter)` (in `standalone` mode especially)

# Issues

- I renamed `"locale" -> "language"` since a locale is something like `"fr-FR"` and a language `"fr"`, and we use `"fr"` as value for the language ; but i can revert these changes and bring back the `"locale"`
- Cannot use `@Input (getter)` to uniformize the code, because the getter is called at any change (including mouse move), then involving too many actions and crash the application
- (Edit: removed) Could we also remove this button ? @aiAdrian  or is it still used for demo purpose on your side ?
image

# Checklist

### DIFF
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -20,7 +20,6 @@ import {NodeService} from "./services/data/node.service";
 export class AppComponent {
   readonly disableBackend = environment.disableBackend;
   readonly version = packageJson.version;
-  readonly locale = localStorage.getItem("locale");
   readonly environmentLabel = environment.label;
   readonly authenticated: Promise<unknown>;
 
@@ -64,6 +63,8 @@ export class AppComponent {
     this.language = language;
     location.reload();
   }
+
+  @Input() language: string;
 
   @Input()
   get netzgrafikDto() {
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -11,6 +11,7 @@ import {NetzgrafikDto} from "./data-structures/business.data.structures";
 import {Operation} from "./models/operation.model";
 import {LabelService} from "./services/data/label.serivce";
 import {NodeService} from "./services/data/node.service";
+import {I18nService} from "./core/i18n/i18n.service";
 
 @Component({
   selector: "sbb-root",
@@ -22,6 +23,7 @@ export class AppComponent {
   readonly version = packageJson.version;
   readonly environmentLabel = environment.label;
   readonly authenticated: Promise<unknown>;
+  protected localStorageLanguage: string;
 
   projectInMenu: Observable<ProjectDto | null>;
 
@@ -45,8 +47,10 @@ export class AppComponent {
               private trainrunSectionService: TrainrunSectionService,
               private nodeService: NodeService,
               private labelService: LabelService,
+              private i18nService: I18nService,
             ) {
-    this.language = localStorage.getItem("i18nLng");
+    this.i18nService.setLanguage();
+    this.localStorageLanguage = localStorage.getItem("i18nLng");
     if (!this.disableBackend) {
       this.authenticated = authService.initialized;
     }
@@ -58,13 +62,14 @@ export class AppComponent {
     }
   }
 
-  changeLanguage(language: string) {
-    localStorage.setItem("i18nLng", language);
-    this.language = language;
-    location.reload();
+  @Input()
+  set language(value: string) {
+    if (value !== this.localStorageLanguage) {
+      localStorage.setItem("i18nLng", value);
+      this.i18nService.setLanguage();
+      this.localStorageLanguage = localStorage.getItem("i18nLng");
+    }
   }
-
-  @Input() language: string;
 
   @Input()
   get netzgrafikDto() {
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -23,7 +23,7 @@ export class AppComponent {
   readonly version = packageJson.version;
   readonly environmentLabel = environment.label;
   readonly authenticated: Promise<unknown>;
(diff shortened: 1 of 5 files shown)


aiAdrian (approved):
Looks good:

**Remove**
image

**Don't remove**
image

This language switch should stay there because the user might want to switch manually the language preference - especially he works for a company with fixed language covernance which is represented in the  IT system. For example, if we have a business system installation, it might depend on the system administrator's settings. For instance, we have German installations, but users might prefer to have another language.

Uriel-Sautron (approved):
LGTM, tested good work.

src/app/app.component.ts
```diff
@@ -58,9 +62,13 @@ export class AppComponent {
     }
   }
 
-  changeLocale(locale: string) {
-    localStorage.setItem("locale", locale);
-    location.reload();
+  @Input()
+  set language(value: string) {
```
emersion, line 66: Would be nice to add a getter which just returns `this.localStorageLanguage`, so that external users can fetch the current language used by NGE as well.

src/app/core/i18n/i18n.service.ts
```diff
@@ -5,41 +5,46 @@ import {loadTranslations} from "@angular/localize";
...
-        localStorage.setItem("locale", this.locale);
+    language: string;
+
+    async setLanguage() {
+      const userLanguage = localStorage.getItem("i18nLng");
+      if (userLanguage && this.allowedLanguages.includes(userLanguage)) {
+        this.language = userLanguage;
       }
-  
+      else {
```
emersion, line 18: Style nit: should be on the same line as the previous one

src/app/core/i18n/i18n.service.ts
```diff
@@ -5,41 +5,46 @@ import {loadTranslations} from "@angular/localize";
...
-      const userLocale = localStorage.getItem("locale");
-  
-      // If the user has a preferred language stored in LocalStorage, use it.
-      if (userLocale && this.allowedLocales.includes(userLocale)) {
-        this.locale = userLocale;
-      } else {
-        localStorage.setItem("locale", this.locale);
+    language: string;
+
+    async setLanguage() {
```
emersion, line 13: Hm, it's a bit awkward to make it the caller's responsibility to set the local storage item, just so that this function can fetch it again. Could we perhaps make it so `setLanguage()` takes an optional `language` argument, and all of the local storage related shenanigans stay inside the i18n service?

src/app/core/i18n/i18n.service.ts
```diff
@@ -5,41 +5,46 @@ import {loadTranslations} from "@angular/localize";
...
       }
-  
+      else {
+        const navigatorLanguage = navigator.language.slice(0, 2);
+        if (this.allowedLanguages.includes(navigatorLanguage)) {
+          this.language = navigatorLanguage;
+        } else {
+          this.language = this.allowedLanguages[0];
+        }
+        localStorage.setItem("i18nLng", this.language);
```
emersion, line 25: If the user hasn't explicitly set a language from the menu, it would be best to leave local storage unset. That way, if the user then changes their preferred language in their web browser setting, NGE will take that into account instead of sticking to the old one.

src/app/app.component.ts
louisgreiner: true !

src/app/core/i18n/i18n.service.ts
louisgreiner: I agree, it is changed, thanks for the spot

src/app/core/i18n/i18n.service.ts
louisgreiner: Done, should be ok now

src/app/core/i18n/i18n.service.ts
```diff
@@ -5,85 +5,95 @@ import {loadTranslations} from "@angular/localize";
...
-      );
-  
-      // Ensure translations are flattened if necessary
-      this.translations = this.flattenTranslations(localeTranslationsModule.default);
-  
-      // Load translations for the current locale at runtime
-      loadTranslations(this.translations);
+export class I18nService {
+  readonly allowedLanguages = ["en", "fr", "de", "it"];
+  private current_language: string = this.getLanguageFromStorage() || this.detectNavigatorLanguage();
```
emersion, line 10: Nit: camelCase is used in NGE instead of snake_case

louisgreiner: oups

emersion (commented):
Apart from these two minor issues, looks good to me, thanks for refactoring!

src/app/core/i18n/i18n.service.ts
```diff
@@ -5,85 +5,95 @@ import {loadTranslations} from "@angular/localize";
...
+
+  get language(): string {
+    return this.current_language;
+  }
+
+  async setLanguage(language?: string) {
+    if (language && this.allowedLanguages.includes(this.language)) {
+      this.setLanguageToStorage(language);
+      this.current_language = language;
+    } else if (!this.allowedLanguages.includes(this.current_language)) {
```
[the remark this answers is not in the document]
emersion: Ah right, good point! Alternatively we could add something like this in `getLanguageFromStorage()`:

```ts
let lang = localStorage.getItem(…);
return this.allowedLanguages.includes(lang) ? lang : null;
```

but the current logic sounds fine to me as well!

emersion (approved):
LGTM :)

PR #394, opened 2024-12-19
Structuring respository by adding folders

NickyDelgado69:
# Description
As part of a university class we were tasked with creating a pull request on an ongoing OSS project with a contribution idea that we worked on

The change hereby simply include adding folders and movind files into them to create a more structured, less packed look for the Repository.

Feel free to proceed however you like with this change as the simple creation of this PR is sufficient to fulfill the task we got to do in class.

I have created the following issue on the matter:

# Issues
(https://github.com/SchweizerischeBundesbahnen/netzgrafik-editor-frontend/issues/375)


louisgreiner:
Hi, thanks for the proposition, my opinion is that it is a bad idea to remove the `.md` files, which should stick to the root of the project (standard), to be displayed on the first page of the GitHub repository.

Even tho it feels like a lot of files, for my use it is not an issue at all. Also, adding folders is a way to potentially make files invisibles. We could think on a way to create a folder for the front-end code part, which would just lead to nest all the code in another folder, which I don't recommend.

However, as I am not a specialist, I summon @emersion, who might has a wider knowledge on the matter.

emersion:
Yeah, I'm sorry, this is not a contribution we can accept. I'd suggest asking first developers of a project whether a change is desirable before making the change. See also the points raised here: https://github.com/OpenRailAssociation/osrd/pull/10131#issuecomment-2554339010

PR #419, opened 2025-03-24
Don't overwrite initial DTO in standalone mode

emersion:
In standalone mode, the DTO might be passed as an HTML element custom attribute. The attribute can be set before NGE is inserted into the DOM like so:

    const sbbRoot = document.createElement('sbb-root');
    sbbRoot.netzgrafikDto = dto;
    document.body.appendChild(sbbRoot);

However, when using the custom attribute this way, the DTO gets overwritten with the default DTO when NetzgrafikApplicationComponent initializes.

Fix this by loading the DTO in AppComponent.ngOnInit instead, only if the user hasn't set the netzgrafikDto custom attribute.

This allows users to set an initial DTO before NGE starts up. Previously, users had to set the DTO after inserting NGE into the DOM and this caused a visible glitch: the default DTO was visible for a split second before the user-provided DTO got loaded.

### DIFF
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -1,4 +1,4 @@
-import {Component, Input, Output} from "@angular/core";
+import {Component, Input, Output, OnInit} from "@angular/core";
 import {AuthService} from "./services/auth/auth.service";
 import {TrainrunService} from "./services/data/trainrun.service";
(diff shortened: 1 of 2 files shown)


aiAdrian (approved):

louisgreiner (approved):
Tested in OSRD, it works very nicely! thx

emersion:
Thank you both for reviewing!

PR #427, opened 2025-04-01
fix: The performance gets improved for big Netzgrafik

aiAdrian:
# Description

# Issues

# Checklist

### DIFF
--- a/src/app/view/editor-main-view/data-views/connections.view.ts
+++ b/src/app/view/editor-main-view/data-views/connections.view.ts
@@ -292,9 +292,12 @@ export class ConnectionsView {
   }
 
   displayConnections(inputConnections: Connection[]) {
-    const connections = inputConnections.filter((c) =>
-      this.filterConnectionsToDisplay(c) &&
-      this.editorView.doCullCheckPositionsInViewport(c.getPath())
+    const connections = inputConnections.filter((c) => {
+        if (this.editorView.doCullCheckPositionsInViewport(c.getPath())) {
+          return this.filterConnectionsToDisplay(c);
+        }
+        return false;
+      }
     );
     const connectionsGroup = this.connectionsGroup
       .selectAll(StaticDomTags.CONNECTION_ROOT_CONTAINER_DOM_REF)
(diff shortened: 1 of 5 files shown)


src/app/view/editor-main-view/data-views/connections.view.ts
```diff
@@ -292,9 +292,12 @@ export class ConnectionsView {
...
   displayConnections(inputConnections: Connection[]) {
-    const connections = inputConnections.filter((c) =>
-      this.filterConnectionsToDisplay(c) &&
-      this.editorView.doCullCheckPositionsInViewport(c.getPath())
+    const connections = inputConnections.filter((c) => {
+        if (this.editorView.doCullCheckPositionsInViewport(c.getPath())) {
+          return this.filterConnectionsToDisplay(c);
+        }
+        return false;
+      }
```
louisgreiner, line 300: If `doCullCheckPositionsInViewport` is a good discriminant, and takes less performance than `filterConnectionsToDisplay`, I'll suggest to implement it that way. It takes less code, and dismiss early cases the way you want. Then `&&` operator ensures that the second operation is not called if the first returns `false` anyway.

```suggestion
const connections = inputConnections.filter((c) =>
      this.editorView.doCullCheckPositionsInViewport(c.getPath()) &&
      this.filterConnectionsToDisplay(c)
```

And the same for the other objects

louisgreiner (commented):
I could not test it right now, but if you can test that the performances are the same with my suggested change, I would approve that way :)

aiAdrian:
Before change:

EditorView.onViewBoxChanged: 508ms 

After change (this fix):

EditorView.onViewBoxChanged: 128ms

SpeedUp 4x

louisgreiner (approved):
Good catch ! Lgtm

louisgreiner:
Can you add "Close https://github.com/SchweizerischeBundesbahnen/netzgrafik-editor-frontend/issues/426" to the body of the issue ? Then, when the PR will be merged, the related issue will be automatically closed.

PR #431, opened 2025-04-03
fix: Performance improvement - when paning the undoSerivce should not…

aiAdrian:
… update the "undo stack"

# Description
If the user wants to pan – that is, only move the view area – the UndoService should not update the stack or check for changes – because only the view area is changed.

# Issues

# Checklist

* [ ] CI is currently green and this is ready for review

### DIFF
--- a/src/app/services/data/undo.service.ts
+++ b/src/app/services/data/undo.service.ts
@@ -66,6 +66,10 @@ export class UndoService implements OnDestroy {
     this.undoRecordingStopped = false;
   }
 
+  getUndoRecording(): boolean {
+    return this.undoRecordingStopped;
+  }
+
   private subscribeSnapshots() {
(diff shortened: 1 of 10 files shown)


aiAdrian:
I am not sure whether the proposed solution is goal-oriented. The problem with quite big Netzgrafik is that the comparision take few computing. But when just pan the viewport this would not be required. 

https://github.com/SchweizerischeBundesbahnen/netzgrafik-editor-frontend/blob/main/src/app/services/data/undo.service.ts#L87

At the moment I see just two path to follow: 
- disable the comparision when panning (the viewport)
- ignore the issue - the generality will be not broken, thus everything works for all kind of operation similar (currently my prefered variant) 

Any other ideas/comments?

aiAdrian:
Will not be merged!

PR #444, opened 2025-05-06
fix: Performance - only render trainrunSection and transition element - which are required

aiAdrian:
# Description
 
 Issue reported by @louisgreiner 
 
 Sarah and Yohan made a discovery about how the train run sections are rendered: for the same section, 4 HTML objects are always rendered (with the same ID, which is surprising for HTML). There is no difference between the 4 objects, except that one of them has a different opacity. Is this normal, or does it serve a special purpose? If not, could it be a potential area for performance improvement to render only one object?
 
image

### DIFF
--- a/documentation/DATA_MODEL_JSON.md
+++ b/documentation/DATA_MODEL_JSON.md
@@ -554,7 +554,7 @@ LinePatternRefs {
 
 ![image](https://github.com/user-attachments/assets/f12d1fe4-cb30-4d9c-95f6-bbe026eae1c0)
 
-This image illustrates the four-layer (level) approach. Start by drawing the background on layer 0, then add levels 1, 2, and 3. Choose the appropriate stroke width and stroke color. Finally, ensure that we always have level 0, which serves as the background event handling area for mouse hovering and clicks (selection). To ensure that mouse events are captured, the stroke opacity for level 0 is set to 0.001 when it is not visible. Rendering styling is controlled via trainrunsections.view.scss 
+The above image illustrates the four-layer (level) approach. Start by drawing the background on layer 0, then add levels 1, 2, and 3. Choose the appropriate stroke width and stroke color. Finally, ensure that we always have level 0, which serves as the background event handling area for mouse hovering and clicks (selection). To ensure that mouse events are captured, the stroke opacity for level 0 is set to 0.001 when it is not visible. Rendering styling is controlled via trainrunsections.view.scss 
 
 
 See also, [TrainrunSectionsView::make4LayerTrainrunSectionLines](https://github.com/search?q=repo%3ASchweizerischeBundesbahnen/netzgrafik-editor-frontend%20make4LayerTrainrunSectionLines&type=code)
--- a/documentation/DATA_MODEL_JSON.md
+++ b/documentation/DATA_MODEL_JSON.md
@@ -551,8 +551,10 @@ LinePatternRefs {
   "120", // -.-.-.  ; unique indentifier
 }
 ```
-The rendering pattern [TrainrunSectionsView::make4LayerTrainrunSectionLines](https://github.com/search?q=repo%3ASchweizerischeBundesbahnen/netzgrafik-editor-frontend%20make4LayerTrainrunSectionLines&type=code)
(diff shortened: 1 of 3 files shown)


aiAdrian:
With this "fix" -> Please test very well - as well also for new frequency which are not in [15, 20, 30, 60, 120] - you might add a new one ~ 240 . insert it directly into the meta data 

---

image

louisgreiner (commented):
Nice spot indeed, however I still don't understand how [this machinery](https://github.com/SchweizerischeBundesbahnen/netzgrafik-editor-frontend/blob/95aad25eddf24b4aaf71fa0a068b0181edf110c3/src/app/view/editor-main-view/data-views/trainrunsections.view.ts#L2484) works... would it be possible to write down a few lines to explain why and how this has been designed and works?

Anyway, after testing, I detected no regression 👍

louisgreiner:
Another point, reported by @Yohh : the `id` attribute of the generated HTML elements currently has the same value for each elements of a section. This might be dangerous to deal with multiple HTML elements that have the same `id`.

Is it for a specific reason? Or could the `id` attributes be changed?

aiAdrian:
> It also raises an issue about `LinePatternRef`.
> 
> I understand that the generation of these HTML elements depend on the style of the line that we want to render, that is why you switched it to the `LinePatternRef` attribute, instead of the `Frequency` attribute: this is a nice fix actually.
> 
> I might open an issue to think on the design of the `LinePatternRef` value, which is shared / combined from the `Trainrun.FrequencyId` and `Trainrun.TimeCategoryId` attributes. It might be easier for understanding if the `LinePatternRef` is a standalone attribute of a trainrun.

Thank you for your comment. We might really need to turn this discussion into a new topic, as LinePatternRef is defined twice in the metadata, which can be confusing. It automatically controls the rendering by blending the two LinePattern. The idea behind it is that we can combine the line patterns from Frequency and TimeCategory, thus having maximum flexibility. However, it does make the rendering somewhat more complex to understand.

Documentation | [Link](https://github.com/SchweizerischeBundesbahnen/netzgrafik-editor-frontend/blob/main/documentation/DATA_MODEL_JSON.md#metadata)

I suggest that we open a new issue here if it needs to be discussed. => Decision? But I think that this discussion should not be held regarding the problem solved by this change, as it is a conceptual issue. Here, it is a filtering problem, meaning that the problem is not the same.

`
TrainrunFrequency::LinePatternRefs {
  "120", // -.-.-.  ; unique indentifier
  "60", // ----- ; unique indentifier
  "30", // ==== ; unique indentifier
  "20", // three lines  ; unique indentifier
  "15", // four lines ; unique indentifier
}

TrainrunTimeCategory::LinePatternRefs {
  TimeCat7_24 = "7/24", // . . . ; unique indentifier
  TimeCatHVZ = "HVZ", // . : . : ; unique indentifier
  TimeZeitweise = "ZEITWEISE", // : - : - ; unique indentifier
}

`

aiAdrian:
> Another point, reported by @Yohh : the `id` attribute of the generated HTML elements currently has the same value for each elements of a section. This might be dangerous to deal with multiple HTML elements that have the same `id`.
> 
> Is it for a specific reason? Or could the `id` attributes be changed?
> 
> If this is not an easy question, we can open a new issue and deal with it later.

Good point. This is also a different topic. :-) 

The ID corresponds to the TrainrunSection.ID and may need to be adjusted through refactoring. If we want to change this, we need to be very careful to ensure that there are no other problems (side effects), as the ID is often used in the code, and the TrainrunSections are accessed using d3.select().... attribute(id) queries, meaning it could potentially become a bigger issue. The ID has nothing to do with the HTML ID, which could be very confusing. The pattern may also be used for other Netzgrafik - objects.

aiAdrian:
> Nice spot indeed, however I still don't understand how [this machinery](https://github.com/SchweizerischeBundesbahnen/netzgrafik-editor-frontend/blob/95aad25eddf24b4aaf71fa0a068b0181edf110c3/src/app/view/editor-main-view/data-views/trainrunsections.view.ts#L2484) works... would it be possible to write down a few lines to explain why and how this has been designed and works?
> 
> Anyway, after testing, I detected no regression 👍

image

louisgreiner (approved):
Wow thanks for the documentation! That's for sure clearer :-)

I wrote an [issue about design of `LinePattern`](https://github.com/SchweizerischeBundesbahnen/netzgrafik-editor-frontend/issues/448), we can continue thinking there.

I understand the issue on `id`s. Maybe we'll discuss it later.

This PR is good to go for me

PR #445, opened 2025-05-07
fix: remove undoService.pushCurrentVersion - call …

aiAdrian:
…when the viewbox gets changed (performance improvement) for zoom/pan

# Description
Comparing a large Netzgrafik to detect changes is time-consuming. When the user pans or zooms the Netzgrafik, an update event is triggered, and the viewbox changes. The update call is required to ensure correct rendering (frustum culling). The changes to the viewbox only update the render area (visualisation). However, neither step changes the data (the Netzgrafik data). Therefore, we don't need to check whether the Netzgrafik has changed and we never push it onto the undo stack. We can just ignore this comparison (step).

These changes improve performance by temporarily disabling the check and pushing the data onto the stack (which would never be necessary without changes).

### DIFF
--- a/src/app/services/data/undo.service.ts
+++ b/src/app/services/data/undo.service.ts
@@ -92,10 +92,11 @@ export class UndoService implements OnDestroy {
   }
 
   public pushCurrentVersion(enforce = false) {
-    this.ignoreNextPushCurrentVersionCall = false;
     if (this.ignoreNextPushCurrentVersionCall && !enforce) {
+      this.ignoreNextPushCurrentVersionCall = false;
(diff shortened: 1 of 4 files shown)


louisgreiner (approved):
For a curious reason, `pushCurrentVersion` is called on left click down, but not on left click down kept pressed ; could this be optimized also? We would like to have the minimum calls of this function.

https://github.com/user-attachments/assets/e0c3838e-d4a7-40c2-abdf-9060c2704df4

Otherwise, after testing, the behavior looks good. The versionning system still works 👍

PR #466, opened 2025-05-22
od: enable svg and png exports

louisgreiner:
Close https://github.com/OpenRailAssociation/osrd/issues/11488

The exported files have not an ideal width and height. This must be reworked, perhaps in another issue.

I would find it cool that the exported files have an adaptive width and height, depending on the size of the elements (either for the network graphic, the space-time chart, or the origin-destination matrix), and not on the viewport (more or less the current behavior)

@aiAdrian what do you think?


src/assets/i18n/it.json
```diff
@@ -436,13 +436,15 @@
         "editor-tools-view-component": {
           "more-functions": "More functions",
           "export": "Export",
-          "export-netzgrafik-as-svg": "Export netzgrafik as SVG",
+          "export-container-as-svg": "Export as SVG: {$container}",
```
shenriotpro, line 439: nit: why not keep the current format?

shenriotpro:
Printing doesn't seem to work for me, possibly because of the default position

aiAdrian:
image

Please have a look how we export the netzgrafik -> the image must have all nodes and text shown. Just change the viewbox i guess.

image
There is a bug in the svg -> viewbox NAN

you have to find out how to make a correct viewbox -> all data are in the svg 

image

src/assets/i18n/it.json
louisgreiner: The previous implementation was actually bugged : this translation was used for all kind of files exports ; 

I either had to create new entries (for Space-time chart and O/D matrix), or do it that way

louisgreiner:
@aiAdrian, I'm afraid the current behavior of the export (for netzgrafik) i bugged.

For this variant
image

I zoom on this part
image

And I try to export:
- netzgrafik:
  - svg: only around the existing train is exported
reticulaire
  - png: only around the existing train is exported
reticulaire
- streckengrafik
  - svg: bugged
graphiqueEspaceTemps
  - png: bugged
graphiqueEspaceTemps
- origin-destination: 
  - svg: not adapted to the matrix size
origineDestination
  - png: not adapted to the matrix size
origineDestination

I will fix for the origin-destination, but the feature is bugged for the 2 other products @aiAdrian

louisgreiner:
.svg and .png now work for origin-destination

Print not yet

src/assets/i18n/it.json
louisgreiner: I wasn't happy with the previous formulation : "Export netzgrafik as SVG". I would prefer "Export **the** netzgrafik as SVG". But articles do not get well translated:
- the network graphic // die Netzgrafik // le réticulaire
- the space-time chart // die StreckenGrafik // le graphique espace-temps
- the origin-destination matrix // die Origin-Destination Matrix // **la** matrice origine-destination

The last "la" makes it harder to translate with 1 translation entry only : I would have needed to take the gender of the word into account also, and this is a pain 😄

shenriotpro: I see, thanks

louisgreiner:
The print feature does not work yet, and is also bugged https://github.com/SchweizerischeBundesbahnen/netzgrafik-editor-frontend/issues/470

We still might want to fix it for O/D matrix

PR #467, opened 2025-05-22
od: make od nodes bold on mouseover

louisgreiner:
With that change, the origin and destination nodes are bold when the mouse is over a cell.

I am not sure on:
- the relevance of this enhancement
- if relevant, the font-size chosen : 12px or less/more?
- if relevant, we could also change the color of the text, or anything else that make the nodes names highlit

What do you think @aiAdrian ?

### DIFF
--- a/src/app/origin-destination/components/origin-destination.component.ts
+++ b/src/app/origin-destination/components/origin-destination.component.ts
@@ -120,13 +120,13 @@ export class OriginDestinationComponent implements OnInit, OnDestroy {
     graphContentGroup
       .append("g")
       .style("pointer-events", "none")
-      .style("font-size", 15)
       .attr("transform", "translate(0, -20)")
       .call(d3.axisBottom(x).tickSize(0))
       .style("user-select", "none")
       .call((g) =>
         g
           .selectAll("text")
+          .attr("data-origin-label", (d: string) => d)
           .style("text-anchor", "start")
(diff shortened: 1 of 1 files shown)


aiAdrian (changes requested):
Very nice idea. But only one bug - if you move the mouse from left to right or from top to bottom e.g. Luzern an then we are going over Luzern, Luzern the bold gets removed. shoud stay bold -

louisgreiner:
I cannot reproduce the bug, can you share it?

shenriotpro:
> I cannot reproduce the bug, can you share it?
> 
>  Enregistrement.de.l.ecran.2025-05-23.095222.mp4

I believe the bug is about the diagonal

src/app/origin-destination/components/origin-destination.component.ts
```diff
@@ -175,11 +179,21 @@ export class OriginDestinationComponent implements OnInit, OnDestroy {
 
     // Three function that change the tooltip when user hover / move / leave a cell
     const mouseover = function (_d) {
```
shenriotpro, line 181: nit: if used, please rename _d to d

shenriotpro (approved):

src/app/origin-destination/components/origin-destination.component.ts
louisgreiner: Done

PR #468, opened 2025-05-22
Lgr/remove pointer events empty cells

louisgreiner:
Currently, when mouse overing empty cells, the tooltip shows this:
image

It gives no information and overload the interface. I suggest removing the tooltips when the cells are empty.

This PR is also a suggestion: I am not sure if we want the disable to tooltip for empty cells

Another solution would be to at least write something else than `undefined`, but I don't have any idea

@aiAdrian


aiAdrian (approved):

shenriotpro:
I think the empty string would be better than `undefined`, I'll open a PR.

shenriotpro (approved):

PR #480, opened 2025-06-11
scale-netzgrafik: fix touchpad gesture with ctrl event listener

louisgreiner:
# Description

Closes https://github.com/SchweizerischeBundesbahnen/netzgrafik-editor-frontend/issues/464

The above issue is a known issue : https://stackoverflow.com/questions/5527601/normalizing-mousewheel-speed-across-browsers

# Issues

`d3.event.ctrlKey` is not detect when track pad is used. The solution is to get this event through an event listener.

# Checklist

* [ x I've read the [Contribution Guidelines](https://github.com/SchweizerischeBundesbahnen/netzgrafik-editor-frontend/blob/main/CONTRIBUTING.md)

### DIFF
--- a/src/app/view/util/svg.mouse.controller.ts
+++ b/src/app/view/util/svg.mouse.controller.ts
@@ -35,19 +35,22 @@ export class SVGMouseController {
   private viewboxIsFixed = false;
 
   private lastMouseEventTimeStamp: number = undefined;
-  private static ctrlKeyPressed = false;
+  private ctrlKeyPressed = false;
 
   constructor(
     private svgName: string,
     private svgMouseControllerObserver: SVGMouseControllerObserver,
     private undoService: UndoService,
-  ) {}
+  ) {
+    this.ctrlKeyDownListener = this.ctrlKeyDownListener.bind(this);
+    this.ctrlKeyUpListener = this.ctrlKeyUpListener.bind(this);
+  }
 
   private ctrlKeyDownListener(e: KeyboardEvent) {
-    if (e.key === "Control") SVGMouseController.ctrlKeyPressed = true;
+    if (e.key === "Control") this.ctrlKeyPressed = true;
   }
   private ctrlKeyUpListener(e: KeyboardEvent) {
-    if (e.key === "Control") SVGMouseController.ctrlKeyPressed = false;
+    if (e.key === "Control") this.ctrlKeyPressed = false;
   }
 
   init(viewboxProperties: ViewboxProperties) {
@@ -351,7 +354,7 @@ export class SVGMouseController {
 
     // We can't use d3.event.ctrlKey because it'll be set to true during
     // pinch-to-zoom touchpad gestures. Use keydown/keyup event listeners instead.
-    if (!SVGMouseController.ctrlKeyPressed) {
+    if (!this.ctrlKeyPressed) {
       // mouse wheel
       if (d3.event.deltaY > 0) {
         this.zoomOut(zoomCenter);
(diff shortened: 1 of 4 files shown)


emersion:
Indeed, there is no way to make the difference between a pinch-to-zoom touchpad gesture and Ctrl+mousewheel using `Event.ctrlKey`.

Alternatively, we could probably install a `keydown` and a `keyup` listener on `window`, and use a variable to indicate whether the Ctrl key is pressed instead of using `Event.ctrlKey`?

src/app/view/util/svg.mouse.controller.ts
```diff
@@ -35,12 +35,21 @@ export class SVGMouseController {
...
+  private static ctrlKeyPressed = false;
 
   constructor(
     private svgName: string,
     private svgMouseControllerObserver: SVGMouseControllerObserver,
     private undoService : UndoService
   ) {
+    // Listens to ctrl key events to differentiate
+    // between pinch-to-zoom touchpad gesture and ctrl + mousewheel
+    window.addEventListener("keydown", (e) => {
```
emersion, line 47: This is done every time a `SVGMouseController` is constructed, would be nice to unregister the callback when it's destroyed.

Also it seems like `init` is the place where other event listeners are set up?

src/app/view/util/svg.mouse.controller.ts
```diff
@@ -35,12 +35,21 @@ export class SVGMouseController {
   private viewboxIsFixed = false;
 
   private lastMouseEventTimeStamp: number = undefined;
+  private static ctrlKeyPressed = false;
```
emersion, line 38: This is a static variable, but the callbacks that update it are per-instance. Maybe this should be per-instance?

emersion (commented):
Otherwise looks pretty good to me!

src/app/view/util/svg.mouse.controller.ts
louisgreiner: True ! fixed

src/app/view/util/svg.mouse.controller.ts
```diff
@@ -35,13 +35,20 @@ export class SVGMouseController {
...
   constructor(
     private svgName: string,
     private svgMouseControllerObserver: SVGMouseControllerObserver,
-    private undoService : UndoService
-  ) {
-  }
+    private undoService: UndoService,
+  ) {}
+
+  private ctrlKeyDownListener = (e: KeyboardEvent) => {
```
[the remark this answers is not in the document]
louisgreiner: Actually, like this it doesn't work, 2 working ways are:
```ts
  private static ctrlKeyPressed = false;
  private ctrlKeyDownListener = (e: KeyboardEvent) => {
    if (e.key === "Control") SVGMouseController.ctrlKeyPressed = true;
  };
  private ctrlKeyUpListener = (e: KeyboardEvent) => {
    if (e.key === "Control") SVGMouseController.ctrlKeyPressed = false;
  };
```
or
```ts
  private ctrlKeyPressed = false;
  private ctrlKeyDownListener = (e: KeyboardEvent) => {
    if (e.key === "Control") SVGMouseController.ctrlKeyPressed = true;
  };
  private ctrlKeyUpListener = (e: KeyboardEvent) => {
    if (e.key === "Control") SVGMouseController.ctrlKeyPressed = false;
  };
```

Otherwise, `this` is not correct, and the ctrl key is not detected anymore. What do you pref ?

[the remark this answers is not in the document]
louisgreiner: I did not know, thanks! done

PR #487, opened 2025-06-26
Sba/enable one way bis

louisgreiner:


aiAdrian (approved):

aiAdrian:
- Filtering : Reset is not working correct for "sym." Onve i reset we should see <-> and -> as it was

louisgreiner:
Indeed, nice catch, I amended the commit with a fix @aiAdrian

PR #518, opened 2025-08-22
fix: Incorrect Handling of One-Way Train Runs in Track Data Assignment

aiAdrian:
# Description

Fix Implemented:

The logic for the checkOneWayOppositeDirection condition was revised to ensure it accurately checks for the correct key format for one-way train runs. This ensures that track data is copied even when the section is in the opposite direction of the defined key.

- [ ] This PR contains a description of the changes I'm making
- [ ] I've read the [Contribution Guidelines](https://github.com/SchweizerischeBundesbahnen/netzgrafik-editor-frontend/blob/main/CONTRIBUTING.md)
- [ ] CI is currently green and this is ready for review


louisgreiner (approved):
LGTM ! Many thanks

Issue #520, opened 2025-08-22
[Bug]: One-way trainrun generates wrong O/D matrix


aiAdrian:
### Preflight Checklist

- [x] I have read the [Contributing Guidelines](https://github.com/SchweizerischeBundesbahnen/netzgrafik-editor-frontend/blob/main/CONTRIBUTING.md) for this project.
- [x] I agree to follow the [Code of Conduct](https://github.com/SchweizerischeBundesbahnen/netzgrafik-editor-frontend/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [x] I have searched the [issue tracker](https://github.com/orgs/sbb-design-systems/projects/1) for an issue that matches the one I want to file, without success.

### Bug type

Functionality

### Which version are you using

2.9.10 (one-way)

### Operating system

all

### Browser / Browser version

all

### What happened?

If we have a network diagram with one-way trains, the O/D matrix calculation is not always accurate / correct. 

Below are three variants:

- V1: There is an ICX train between A and B and a REX train that operates between B and C. All trains run hourly and make a round trip (to and from).
- V2: The ICX operates as in V1, but the REX train only runs in the direction from B to C.
- V3: The ICX operates as in V1, but the REX train only runs in the direction from C to B.

Below you will find the three variants as a netzgrafik and corresponding O/D matrix – each as screenshots:

| Variant | Netzgrafik | O/D-Matrix |
|----------|------------|------------|
| V1       |  <img width="527" height="483" alt="Image" src="https://github.com/user-attachments/assets/90880115-1093-4de8-8877-e72e93656317" /> | <img width="202" height="181" alt="Image" src="https://github.com/user-attachments/assets/6254ed5b-4ae3-42a2-abd0-f237ce32fb47" />  |
| V2       |  <img width="568" height="524" alt="Image" src="https://github.com/user-attachments/assets/9e6bf835-b990-40b6-8c1c-b965afce1192" /> | <img width="295" height="315" alt="Image" src="https://github.com/user-attachments/assets/d58aefcd-6241-4550-9be2-dd5857fc15e8" /> |
| V3       |  <img width="585" height="535" alt="Image" src="https://github.com/user-attachments/assets/379c7387-4773-46c1-8e46-b0002fe61611" /> | <img width="321" height="330" alt="Image" src="https://github.com/user-attachments/assets/8a208f64-6da3-4ad5-a092-4137c8e9e4c3" /> |

[...]

PR #521, opened 2025-08-25
one-way: fix wrong origin-destination matrix calculation

louisgreiner:
Closes https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/issues/520


aiAdrian (changes requested):
As in the comments documented there are still major problems.

louisgreiner:
After further testing, the previous solution seems to be insufficient.

The solution proposed by @shenriotpro seems to work nice. If you don't find any more edge cases, we could go on that fix and think on refactoring the algorithm using the source -> target paradigm instead later.

@aiAdrian

aiAdrian:
# Netzgrafik and O/D Matrix
image

image


# Export O/D Matrix to csv 
[originDestination.csv](https://github.com/user-attachments/files/22515675/originDestination.csv)

image


# Data 
[netzgrafik_test.json](https://github.com/user-attachments/files/22515645/netzgrafik_test.json)

PR #534, opened 2025-09-17
fix: Highlight entire trainrun when hovering a trainrun section line

aiAdrian:
# Description

When the user is moving the mouse pointer over a trainrun section, the trainrun sections stroke-width gets wider. 

# Issues

# Checklist

### DIFF
--- a/src/app/view/editor-main-view/data-views/trainrunsections.view.edgeline.scss
+++ b/src/app/view/editor-main-view/data-views/trainrunsections.view.edgeline.scss
@@ -76,57 +76,57 @@
 }
 
 /// HOVER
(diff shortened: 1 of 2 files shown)


louisgreiner (commented):
The hover's behavior is a bit shaky and does not work so well, could it be fixed in this PR as well?

https://github.com/user-attachments/assets/7c597019-23b2-4cc7-a6d3-8b1ec5d3789d

The idea is nice, but I wonder if we could also highlight not only the hovered section, but the whole trainrun at the same time? What do you think? The business on our side seems to prefer that idea (but we should try first)

aiAdrian:
fix: highlight entire trainrun when hovering a trainrun section line

PR #537, opened 2025-09-23
front: restore disbabled backend on notes

Caracol3:
Restoring the notes feature for standalone environnement.


louisgreiner:
This PR re-enables notes in standalone mode. It was an error to disable them in this repository, and it should have been done on OSRD side, you're right. So we're restauring them, to be able to use them [here](https://openrailassociation.github.io/netzgrafik-editor-frontend/) for instance.

(we are actually working in OSRD to work with the notes yes)

PR #540, opened 2025-09-23
470 bug export features bugged png svg

aiAdrian:
# Description

### 🛠 Inline CSS for SVG PNG Export

I located the problem. It's the classic CSS style loss problem with saveSvgAsPng when exporting, especially if your SVG relies on external stylesheets or scoped component styles. The root cause: saveSvgAsPng only serializes inline styles by default, so anything coming from external CSS or Angular’s component encapsulation can get dropped.

This approach helps preserve visual fidelity, especially for dynamic styles such as media queries, fonts, and layout-specific rules.

### ⚠️ Disclaimer

This solution is still **in progress**. While it improves style retention, performance is currently **very poor** esspecially for Graphical Timetable — the inlining process takes too long for large or complex SVGs. Further optimization is needed before it can be used in production.

- Netzgrafik export : ok.
- O/D Matrix export: ok.
- Graphical timetable export: performance poor.
 

```typescript
const svgElement = document.getElementById("mySvg");
const allElements = svgElement.querySelectorAll("*");

allElements.forEach(el => {
  const computedStyle = window.getComputedStyle(el);
  const styleString = Array.from(computedStyle)
    .map(key => `${key}:${computedStyle.getPropertyValue(key)};`)
    .join(" ");
  el.setAttribute("style", styleString);
});
```
# Summary

### Export Graphical Timetable (Streckengrafik)
Users can export the Graphical Timetable, but they should expect a delay. The export preparation—especially styling, layout, and rendering—takes considerable time. However, the final result is acceptable.  
**Workaround:** Users can just take a screenshot of the timetable take  

### Export Origin/Destination Matrix
Works well and is quite fast.

### Export Netzgrafik
Works well and is quite fast.

## Conclusion
Since the export of the Netzgrafik and Origin/Destination Matrix is essential and users have no alternative, this fix is sufficient to merge.  
The reported issue with exporting the Graphical Timetable (Streckengrafik) is a minor problem. It would be nice to resolve it, but for milestone **2.10**, it is not a critical requirement.

---- 
The https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/issues/542 address the problem with the Graphical Timetable (Streckengrafik) export into png/svg

--- 
| Before fix | After fix |
|---|---| 
|netzgrafik | netzgrafik |
| originDestination|  originDestination |
| streckengrafik| streckengrafik |

### DIFF
--- a/src/app/streckengrafik/components/streckengrafik.component.html
+++ b/src/app/streckengrafik/components/streckengrafik.component.html
@@ -124,126 +124,135 @@
         <!---   new line  --->
         <div class="footer_1 background"></div>
         <div class="footer_2 background">
-          <button
-            type="button"
-            [class]="getZoomButtonClassTag('SimpleButton', 41.6)"
-            title="1/4h"
-            (click)="onFixZoom(41.6)"
-          >
-            15'
-          </button>
-          <button
-            type="button"
-            [class]="getZoomButtonClassTag('SimpleButton', 20.8)"
-            title="1/2h"
-            (click)="onFixZoom(20.8)"
-          >
-            30'
-          </button>
-
-          <button
-            type="button"
-            [class]="getZoomButtonClassTag('SimpleButton', 10.4)"
-            title="1h"
-            (click)="onFixZoom(10.4)"
-          >
-            1h
-          </button>
-          <button
-            type="button"
-            [class]="getZoomButtonClassTag('SimpleButton', 5.2)"
-            title="2h"
-            (click)="onFixZoom(5.2)"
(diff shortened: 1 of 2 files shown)


louisgreiner:
I'm using Firefox, and the zoom level was 100% :/

louisgreiner (approved):
Tested, it looks very good ! Not sure if the code should be re-thought, especially `prepareStyleForExport()`, but this can be done in a later step. The fact that the exports are now fixed is a very good first step :) Thanks

Issue #488, opened 2025-06-26
[Bug]: Space-time chart wrong y-axis


louisgreiner:
### Preflight Checklist

- [x] I have read the [Contributing Guidelines](https://github.com/SchweizerischeBundesbahnen/netzgrafik-editor-frontend/blob/main/CONTRIBUTING.md) for this project.
- [x] I agree to follow the [Code of Conduct](https://github.com/SchweizerischeBundesbahnen/netzgrafik-editor-frontend/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [x] I have searched the [issue tracker](https://github.com/orgs/sbb-design-systems/projects/1) for an issue that matches the one I want to file, without success.

### Bug type

Functionality

### Which version are you using

2.9.29

### Input mode

None

### What happened?

![Image](https://github.com/user-attachments/assets/b5becb05-4de7-4add-907d-9efcbab28bd1)

![Image](https://github.com/user-attachments/assets/2c157306-6b62-42cf-b989-6dc5b28711f9)

The y-axis shows that the train starts around 40', but it is set as 0' in the network graphic

### Steps to reproduce the issue

[reticulaire(1).json](https://github.com/user-attachments/files/20923415/reticulaire.1.json)

### Relevant log output

```shell

```

aiAdrian:
Retest required!!!

louisgreiner:
This is really weird, this is no even a cache issue, I tried also on a cacheless browser

https://github.com/user-attachments/assets/9b34c500-ad66-4970-aaf5-0803418888bd

Could it be an issue on Firefox?

aiAdrian:
I can reproduce this problem - only for firefox :-)

PR #548, opened 2025-10-01
fix: Firefox rendering issue fixed – styling slightly changed

aiAdrian:
# Description

# Issues

# Checklist

### DIFF
--- a/src/app/streckengrafik/components/slider/time-slider/time-slider.component.scss
+++ b/src/app/streckengrafik/components/slider/time-slider/time-slider.component.scss
@@ -36,27 +36,28 @@
 
 .TimeSliderComponent.HorizontalLine {
   stroke: var(--NODE_TEXT_FOCUS);
-  stroke-width: 0.25px;
+  stroke-width: 1px;
+  stroke-dasharray: 1 1;
 }
 
 .TimeSliderComponent.HorizontalLine.Hour.Five {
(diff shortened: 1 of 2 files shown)


louisgreiner:
I just tested:

| Browser | Before fix  | After fix    |
| --- | --- | --- |
| Firefox |  image | image |
| Edge | image | image |

The dashes now are displayed on all environments, nice.
I think I prefered the "Before" version on Edge, the dashes looked better; but I wonder why they did not appeared on Firefox. If this is the solution to make Firefox have the dashes, I'm ok with that.

Also, this does not fix the [bug on y-axis values not coherent with the trainruns](https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/issues/488#issuecomment-3356373775)

louisgreiner:
🥲 I don't know if this is the issue, I have an older version. Still this is very strange that the behaviors are so different...

image

Also, when zooming in / out, the left dashes disappear as well (this could be done another time of course)

aiAdrian:
> 🥲 I don't know if this is the issue, I have an older version. Still this is very strange that the behaviors are so different...
> 
> image
> Also, when zooming in / out, the left dashes disappear as well (this could be done another time of course)

I try to install same version of firefox :-)

louisgreiner (approved):

PR #573, opened 2025-10-16
Aeg/performance compute shortest paths parallel preparation

aiAdrian:
# Description
It would be highly beneficial to process this in parallel using 2 to 4 threads.
By applying Promise-based chunking and await logic, we could significantly reduce total runtime.
Especially for large network graphs, sequential pathfinding can become a major bottleneck.
 
Currently, pathfinding runs sequentially for each origin node and for big netzgrafik it takes quite a lot of time!

```typescript 
   // In theory we could parallelize the pathfindings, but the overhead might be too big.
    const res = new Map<string, [number, number]>();
    odNodes.forEach((origin) => {
      computeShortestPaths(origin.getId(), neighbors, vertices, tsSuccessor, cachedKey).forEach(
        (value, key) => {
          res.set([origin.getId(), key].join(","), value);
        },
      );
    });
```

If we can make the `computeShortestPaths` as well async then the following code should be executable in parallel. You can just add a new method ` computeBatchShortestPaths` to `OriginDestinationService` : 

```typescript
async computeBatchShortestPaths(
    odNodes: Node[],
    neighbors: Map<string, [Vertex, number][]>,
    vertices: Vertex[],
    tsSuccessor: Map<number, number>,
    cachedKey: Map<Vertex, string>,
    res: Map<string, [number, number]>,
  ): Promise<void> {
    const numThreads = 1;
    const chunkSize = Math.ceil(odNodes.length / numThreads);
    const allChunks: [number, number][] = [];

    for (let i = 0; i < odNodes.length; i += chunkSize) {
      allChunks.push([i, Math.min(i + chunkSize, odNodes.length)]);
    }

    const threads: Promise<void>[] = [];

    for (const [start, end] of allChunks) {
      const chunk = odNodes.slice(start, end);

      const thread = new Promise<void>((resolve) => {
        for (const origin of chunk) {
          const results = computeShortestPaths(
            origin.getId(),
            neighbors,
            vertices,
            tsSuccessor,
            cachedKey,
          );

          results.forEach((value, key) => {
            res.set([origin.getId(), key].join(","), value);
          });
        }

        console.log("done:", start, end);
        resolve();
      });

      threads.push(thread);
    }

    await Promise.all(threads);
  }

```

In the `originDestinationData(): OriginDestination[] { ... } ` function (method) we have to replace:

This code snipped 
```typescript
    odNodes.forEach((origin) => {
      computeShortestPaths(origin.getId(), neighbors, vertices, tsSuccessor, cachedKey).forEach(
        (value, key) => {
          res.set([origin.getId(), key].join(","), value);
        },
      );
    });
``` 
with 
 ```typescript
    this.computeBatchShortestPaths(odNodes, neighbors, vertices, tsSuccessor, cachedKey, res);
```

# Issues

 

# Checklist

### DIFF
--- a/src/app/services/analytics/origin-destination/components/origin-destination.service.ts
+++ b/src/app/services/analytics/origin-destination/components/origin-destination.service.ts
@@ -56,7 +56,7 @@ export class OriginDestinationService {
     cachedKey: Map<Vertex, string>,
     res: Map<string, [number, number]>,
   ): Promise<void> {
-    const numThreads = 1;
+    const numThreads = 4;
     const chunkSize = Math.ceil(odNodes.length / numThreads);
     const allChunks: [number, number][] = [];
 
--- a/src/app/services/analytics/origin-destination/components/origin-destination.service.ts
+++ b/src/app/services/analytics/origin-destination/components/origin-destination.service.ts
@@ -10,7 +10,7 @@ import {
   topoSort,
   Vertex,
 } from "src/app/view/util/origin-destination-graph";
-import { M } from "@angular/cdk/keycodes";
+import {M} from "@angular/cdk/keycodes";
 
 // Computed values for an origin/destination pair.
 export type OriginDestination = {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/app/services/analytics/origin-destination/components/origin-destination.service.ts [hunk 1]
```diff
@@ -48,6 +49,148 @@ export class OriginDestinationService {
     return selectedNodes.length > 0 ? selectedNodes : this.nodeService.getVisibleNodes();
   }
 
+  // Given a graph (adjacency list), and vertices in topological order, return the shortest paths (and connections)
+  // from a given node to other nodes.
+  async computeShortestPaths(
```
aiAdrian: Just a copy of the original code but declared as async.

src/app/services/analytics/origin-destination/components/origin-destination.service.ts
```diff
@@ -48,6 +49,148 @@ export class OriginDestinationService {
...
+        res.set(vertex.nodeId, cachedDistGetKey);
+      }
+      const neighs = neighbors.get(key);
+      if (neighs === undefined || cachedDistGetKey === undefined) {
+        return;
+      }
+      // The shortest path from the start node to this vertex is a shortest path from the start node to a neighbor
+      // plus the weight of the edge connecting the neighbor to this vertex.
+      //neighs.forEach(([neighbor, weight]) => {
+      for (let idx = 0; idx < neighs.length; idx++) {
```
aiAdrian: forEach replaced

PR #591, opened 2025-10-27
Fix: One way train run arrows are drawn outside of lines when moving node

aiAdrian:
# Description

I made two changes. **The 1st one fixes the issue,** the second one is a very small change but has an impact to the rendering overall performance. It's shows how we can enable/disable objects for rendering based on a switch .... 

- When dragging a node, the train route section is displayed very abstractly—just as a line. As soon as the node is dragged (finished drag), the full detail rendering is  (re-)applied. This ensures high drag performance and prevents elements that have not yet been fully updated (recalculated) from being displayed incorrectly. This means that not everything always needs to be fully updated when dragging, which further improves performance.

- Add filter to groupLinesEnter: render arrows only for non-round trips

# Issues

# Checklist

### DIFF
--- a/src/app/view/editor-main-view/data-views/trainrunsections.view.ts
+++ b/src/app/view/editor-main-view/data-views/trainrunsections.view.ts
@@ -1058,7 +1058,6 @@ export class TrainrunSectionsView {
         .filter((d: TrainrunSectionViewObject) => !d.trainrunSection.getTrainrun().isRoundTrip())
         .append(StaticDomTags.EDGE_LINE_ARROW_SVG)
         .attr("d", "M-4,-5L2,0L-4,5Z")
-        })
         .attr("transform", (d: TrainrunSectionViewObject) =>
           this.translateAndRotateArrow(d.trainrunSection, arrowType),
         )


src/app/view/editor-main-view/data-views/trainrunsections.view.ts
```diff
@@ -1055,6 +1055,7 @@ export class TrainrunSectionsView {
   ) {
     (["BEGINNING_ARROW", "ENDING_ARROW"] as const).forEach((arrowType) => {
       groupLinesEnter
+        .filter((d: TrainrunSectionViewObject) => !d.trainrunSection.getTrainrun().isRoundTrip())
         .append(StaticDomTags.EDGE_LINE_ARROW_SVG)
         .attr("d", (d: TrainrunSectionViewObject) => {
           return d.trainrunSection.getTrainrun().isRoundTrip() ? "" : "M-4,-5L2,0L-4,5Z";
```
louisgreiner: Then we can even do this

```suggestion
        .filter((d: TrainrunSectionViewObject) => !d.trainrunSection.getTrainrun().isRoundTrip())
        .append(StaticDomTags.EDGE_LINE_ARROW_SVG)
        .attr("d", "M-4,-5L2,0L-4,5Z")
```

louisgreiner (approved):
👌

PR #554, opened 2025-10-07
feat: modify node display

Math-R:
# Description

This PR implements the **collapsed nodes** functionality for the Netzgrafik editor, allowing intermediate nodes to be hidden while maintaining proper visual representation of train routes.

### 1. Node Model Enhancement
- Added `isCollapsed` field to Node model with getter/setter methods
- Integrated collapsed state into node serialization/deserialization
- Maintains backward compatibility with existing data

### 2. Visual Filtering System  
- **NodesView**: Filter out collapsed nodes from visual display
- **ConnectionsView**: Hide connections associated with collapsed nodes
- **TransitionsView**: Hide transitions for collapsed nodes
- Collapsed nodes become completely invisible in the editor

### 3. Intelligent Section Grouping
- **TrainrunSectionService.groupTrainrunSectionsIntoChains()**: Algorithm to group sections with collapsed intermediate nodes
- Automatically detects chains like A→B→C→D where B,C are collapsed
- Creates grouped objects with proper start/end node references

### 4. Direct Path Calculation
- **TrainrunSectionsView.createViewObjectForCollapsedChain()**: Creates direct visual paths A→D bypassing collapsed nodes
- Automatic routing that circumvents collapsed intermediate nodes  
- Proper port calculation for direct connections

## User Experience
- Nodes marked as `collapsed` become invisible
- Train routes A→B→C→D (with B,C collapsed) display as direct A→D connections
- Interface remains coherent and schedules are preserved
- No impact on existing data or workflows

## Technical Implementation
- Zero breaking changes to existing data structures
- Clean separation between data model and visual representation
- Proper TypeScript typing throughout

# Issues

Related to feature request for collapsed nodes functionality to simplify complex network display.

# Checklist

- [x] This PR contains a description of the changes I'm making
- [x] I've read the [Contribution Guidelines](https://github.com/SchweizerischeBundesbahnen/netzgrafik-editor-frontend/blob/main/CONTRIBUTING.md)
- [x] **Core functionality implemented and tested**: Node filtering, section grouping, path calculation working
- [ ] I've added tests for changes or features I've introduced *(comprehensive tests planned for follow-up PR)*
- [ ] I documented any high-level concepts I'm introducing in `documentation/` *(documentation to be added in follow-up)*
- [x] **TypeScript compilation clean**: No compilation errors
- [x] **Backward compatibility preserved**: No breaking changes to existing data
- [x] **Ready for functional testing**: Core collapsed nodes feature fully operational

## Next Steps (Future PRs)
[...]

PR #602, opened 2025-10-31
Fixes for collapsed nodes support in main editor view

emersion:
_See individual commits._

First two commits address https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/pull/554#discussion_r2452370412

Rest starts addressing the last paragraph of https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/pull/554#pullrequestreview-3350620157

Some more follow-up cleanups haven't been written yet.

Note, each time we access `TrainrunSectionViewObject.trainrunSections[0]` needs to be re-evaluated to figure out whether we really should grab the first section, or the last one, or compute something over all sections (e.g. travel time).

### DIFF
--- a/src/app/view/editor-main-view/data-views/trainrunsections.view.ts
+++ b/src/app/view/editor-main-view/data-views/trainrunsections.view.ts
@@ -2910,14 +2910,12 @@ export class TrainrunSectionsView {
     textElement: TrainrunSectionText,
     editorView: EditorView,
   ) {
-    const trainrunSection = viewObject.trainrunSections[0];
-
     switch (textElement) {
       case TrainrunSectionText.SourceDeparture:
       case TrainrunSectionText.SourceArrival:
         // For source times, use the original section times (should be from non-collapsed node)
         return TrainrunSectionsView.getTrainrunSectionValueToShow(
-          trainrunSection,
+          viewObject.trainrunSections[0],
           textElement,
           editorView,
         );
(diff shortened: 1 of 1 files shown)


src/app/services/data/trainrunsection.service.ts
```diff
@@ -1486,10 +1486,19 @@ export class TrainrunSectionService implements OnDestroy {
...
+      const backwardIterator = this.trainrunService.getBackwardIterator(
+        section.getTargetNode(),
+        section,
+      );
+      while (backwardIterator.hasNext() && backwardIterator.current().node.getIsCollapsed()) {
+        backwardIterator.next();
+      }
+      const startNode = backwardIterator.current().node;
+      const startSection = backwardIterator.current().trainrunSection;
+
```
louisgreiner: Maybe we can store this in a helper function, next to `TrainrunService.getFirstTrainrunSection()`. But I feel like this would also embed a wider refacto around `TrainrunService.ExpandedTrainrunIterator`. Ignore this and we deal with it later. Do you agree @Math-R ?

Math-R (approved):
Thanks for this work. It makes the code cleaner and clearer.

Nit: For `getCollapsedChainValueToShow`, personally, I would have kept the function as is. Because its signature allows for self-documentation of this part of the code.

src/app/services/data/trainrunsection.service.ts
emersion: Yeah, I considered extracting this to a helper function, but didn't find a good spot/name for it. Let's leave this for later!

PR #641, opened 2025-11-24
import: set node position before calling replaceIntermediateStopWithNode()

emersion:
replaceIntermediateStopWithNode() recomputes a bunch of transition position stuff. At import time, tt was working with incorrect node positions, routing everything to the (0, 0) position. This causes transition positions to be mis-rendered.

Instead, set the node position at initialization, before calling replaceIntermediateStopWithNode(). That way the function can work with the final node position. As a bonus, the "new node" event is fired with the correct node position.

Fixes: 435f1c66b8ed ("import: handle importing old files with numberOfStops")


louisgreiner (approved):
Nice!

A test has failed tho

emersion:
I think this bug is unrelated. The computed port positions depend on collapsed nodes. We decided to delay fixing this after @Math-R's PR is merged.

PR #660, opened 2025-12-01
Prevent spurious update event on note creation

emersion:
When a note is created, moveNoteToFront() is called. This triggers an "update" event even when the note is already at the top of the stack.

Make it so moveNoteToFront() is a no-op when the note is already at the top.


aiAdrian (approved):
Excellent observation and nice fix !

PR #687, opened 2025-12-11
fix: emit trainrun create after creating its sections when duplicating

Synar:
When duplicating a trainrun, the event to create the trainrun was emitted before the train sections were created. At least in osrd, this created a bug, as osrd can not create a trainrun with no train sections.

### DIFF
--- a/src/app/services/data/trainrun.service.ts
+++ b/src/app/services/data/trainrun.service.ts
@@ -378,6 +378,7 @@ export class TrainrunService {
     newTrainrun.select();
     this.nodeService.transitionsUpdated();
     this.trainrunsUpdated();
+    this.operation.emit(new TrainrunOperation(OperationType.create, newTrainrun));
   }
 
   combineTwoTrainruns(node: Node, port1: Port, port2: Port) {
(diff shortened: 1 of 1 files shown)


src/app/services/data/trainrun.service.ts
```diff
@@ -366,6 +366,7 @@ export class TrainrunService {
       }
       iterator.current().trainrunSection.setTrainrun(newTrainrun);
     }
+    this.operation.emit(new TrainrunOperation(OperationType.create, newTrainrun));
```
louisgreiner: Nite: would is be better to move this call after `    newTrainrun.select();` a few lines below?

Synar: I thought it would be best to do it as soon as the sections are created, to avoid separating it too much from the trainrun creation. I don't know what the implications would be here though, so I will trust you on this. ^^

louisgreiner: I may be wrong and it mostly depends of the context, but the emition should always occur at the very end if possible, otherwise the event can be emitted and an error can occur later in the code

louisgreiner:
If you rebase on `main`, the CI should tick green ✅

src/app/services/data/trainrun.service.ts
emersion: I think we should at least emit the event after `checkAndFixMissingTransitions()`, since that function may update the DTO. (We walk the trainrun section graph via transitions when handling events on the OSRD side.)

PR #688, opened 2025-12-11
chore: git ignore .vscode

Synar:
Some parts of .vscode were not git ignored. However, none of these were actually committed. Additionally, such a folder tends to vary wildly between different developers, and should probably not be committed.


aiAdrian (approved):

emersion:
Since we already have editor-specific rules in there, I think it makes sense to merge this.

PR #691, opened 2025-12-12
refactor: rename propagateTimeAlongTrainrunFixStartTrainrunSection()

emersion:
This function doesn't propagate nor fix anything, it merely walks along a trainrun section until it finds a particular node.


aiAdrian (approved):
LGTM

PR #695, opened 2025-12-12
refactor: drop TrainrunDialogParameter.nodesOrdered

emersion:
_See individual commits._

### DIFF
--- a/src/app/perlenkette/perlenkette-section/perlenkette-section.component.ts
+++ b/src/app/perlenkette/perlenkette-section/perlenkette-section.component.ts
@@ -93,8 +93,6 @@ export class PerlenketteSectionComponent implements OnInit, AfterContentInit, On
     travelTimeLock: false,
   };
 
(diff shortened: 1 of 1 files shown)


emersion:
Have you seen the first commit's description? It reads:

> TrainrunDialogParameter.setNodesOrdered() was never called, so this
> was always an empty array.

`nodesOrder` never had any value other than `[]`, an empty array.

In the dialog, we want to work with all non-stop nodes neighboring the trainrun section. This is handled here:

https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/blob/710723e88b40cad12df81a1f18a696b4f19dd9e2/src/app/services/util/trainrunsection.helper.ts#L306

Do you want me to provide more context? I would be happy to write up a more detailed explanation if the above isn't enough.

Thanks for reviewing!

aiAdrian (approved):
Nice spot - Thank you for the PR. It makes the code more clean

PR #700, opened 2025-12-18
Lgr/fix port assignation

louisgreiner:
Only the last commit has to be reviewed, other are from other branches


louisgreiner:
Re-opened https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/pull/732

PR #702, opened 2025-12-19
fix: address XSS vulnerability with note text

emersion:
Notes can contain HTML in their body text. The HTML was directly piped into the DOM in 3 spots, resulting in an XSS vulnerability. The XSS vulnerability can be used to e.g. perform authenticated API calls on behalf of another user by making a privileged user open a shared variant by sending them a link.

Here are the 3 spots where the note text is used:

- In the notes view, when displaying the body text in the editor main view's SVG with d3. The d3 html() method is used.
- In the notes view, when computing the size of the body text. The innerHTML field is set.
- In the note edit dialog, when passing the text to ngx-editor. ngx-editor doesn't perform any sanitization on its own.

Fix all of these by sanitizing unsafe HTML strings before using them. A new Note.getSanitizedText() method is introduced for this purpose.

Unfortunately we cannot leverage Angular's DomSanitizer.sanitize() method, because it's not flexible enough for us: it disallows style attributes, which are used for text color. There is no way to customize this behavior.

Instead, use the standard [HTML Sanitizer API][1] if available. Safari doesn't yet implement it, so keep a DOMPurify-based fallback in place ([Webkit ticket](https://bugs.webkit.org/show_bug.cgi?id=303808)). We'll be able to drop this fallback when all browsers have the standard API available.

Here is a PoC for the XSS vulnerability: 
[networkGraphic(16).json](https://github.com/user-attachments/files/24254633/networkGraphic.16.json)

[1]: https://developer.mozilla.org/en-US/docs/Web/API/HTML_Sanitizer_API

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -23,6 +23,7 @@
         "angular-oauth2-oidc": "^15.0.1",
         "angular-server-side-configuration": "^18.2.0",
         "d3": "^5.16.0",
+        "dompurify": "^3.3.3",
         "extract-i18n": "^1.0.0",
         "ngx-editor": "^15.0.1",
         "papaparse": "^5.3.2",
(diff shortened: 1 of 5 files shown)


aiAdrian (changes requested):
The proposed change will no longer provide the same behaviour... Colours are no longer be visible in the note dialog.


## Before change 
image


## After change (this PR) 
image

emersion:
Indeed, good catch!

Hm, this is going to be tricker to fix… It doesn't seem like there is a way to convince `DomSanitizer` to allow styles…

emersion:
Sorry for the wait! I've updated this PR with another approach which should fix the colors issue.

aiAdrian (approved):
Thank you

DucNg (approved):

SarahBellaha (approved):
Lgtm

emersion:
Rebased against `main` to fix `package-lock.json` conflicts.

PR #704, opened 2025-12-19
fix: Lock symbol (svg) background rendering issue pearls view

aiAdrian:
# Description

This is just a very small fix. The background color was not set for the side-panel (pearls view) which cause a rendering issue. The rendering issue was only seen on screens with bad contrast adjustments.

# Issues

# Checklist


emersion (commented):
Hm, I think like the sidebar background color is fine, and the lock icon background color should be fixed instead?

Before the patch:

out

After the patch:

out

PR #708, opened 2025-12-22
Strict checks on null values for models directory and its dependencies

### LINKED ISSUE #645: [Feature request]: Add StrictNullCheck constraint to configuration
### Preflight Checklist

- [x] I have read the [Contributing Guidelines](https://github.com/SchweizerischeBundesbahnen/netzgrafik-editor-frontend/blob/main/CONTRIBUTING.md) for this project.
- [x] I agree to follow the [Code of Conduct](https://github.com/SchweizerischeBundesbahnen/netzgrafik-editor-frontend/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [x] I have searched the [issue tracker](https://github.com/orgs/sbb-design-systems/projects/1) for an issue that matches the one I want to file, without success.

### Request type

Request for enhancement of a component

### Functionality

# Description

The goal is to add the "strictNullChecks": true to the config file and get it to work fine with every part of the project.

>     "useDefineForClassFields": false,
>     "strictNullChecks": true

To do so, we can work on mutiple branches focusing on a specific directory and get the files in it to comply with the rule, one by one.

Using `validate:ts` we can check only the directory we're interested into :

### Example

`npm run validate:ts -- --strictNullChecks src/app/utils/*.ts`

This is a suggestion to make the code base more safe. Currently, many issues can happen because of this.

This needs to fix many many things in the code (see CI or run `npm run build`) before this can be merged (CI ticks green), the config can be enabled only on some folders first to split the work in smaller pieces.

# Issues

- https://github.com/OpenRailAssociation/osrd/issues/13531

# Checklist

- [ ] This PR contains a description of the changes I'm making
- [ ] I've read the [Contribution Guidelines](https://github.com/SchweizerischeBundesbahnen/netzgrafik-editor-frontend/blob/main/CONTRIBUTING.md)
- [ ] CI is currently green and this is ready for review

### Link to design proposal file

_No response_

Pivouane:
# Description

Handle null values warnings and errors for the models directory and its dependencies

# Issues

#645 

# Checklist

- [ ] This PR contains a description of the changes I'm making
- [ ] I've read the [Contribution Guidelines](https://github.com/SchweizerischeBundesbahnen/netzgrafik-editor-frontend/blob/main/CONTRIBUTING.md)
- [ ] CI is currently green and this is ready for review


src/app/data-structures/business.data.structures.ts
emersion, line 1: Could we use `undefined` for the DTO structures rather than `null`? When a JSON file is missing a field (because it's been generated by an older NGE), the field will be `undefined`, and `field === null` is false.

See also: https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/pull/805#discussion_r2930080297

src/app/data-structures/technical.data.structures.ts
emersion, line 1: Ditto

tsconfig.strictNullChecks.json
```diff
@@ -6,6 +6,11 @@
     "strictNullChecks": true
   },
   "files": [],
-  "include": ["src/app/view/themes/**/*.ts"],
+  "include": [
+    "src/app/view/themes/**/*.ts",
+    "src/app/data-structures/**/*.ts",
+    "src/polyfills.ts",
+    "src/app/models/**/*.ts"
```
emersion, line 13: Could we also add `src/app/services/util/trainrunsection.routing.ts` to this list?

src/app/models/filterSettings.model.ts
```diff
@@ -73,10 +73,10 @@ export class FilterSetting {
...
       filterConnections: true,
       filterShowNonStopTime: true,
-      filterTrainrunCategory: null,
-      filterTrainrunFrequency: null,
-      filterTrainrunTimeCategory: null,
-      filterDirection: null,
+      filterTrainrunCategory: [],
+      filterTrainrunFrequency: [],
+      filterTrainrunTimeCategory: [],
+      filterDirection: [],
```
emersion, line 89: It seems like `[]` now means "use the default", ie. "use all categories/freqs/etc"? It seems like a breaking change to me: before this PR, `[]` meant "use none of the categories".

In other words: when de-selecting all trainrun categories in the filter sidebar, exporting and re-importing the network graph as JSON, I think this PR makes it so all trainrun categories are selected again.

src/app/models/node.model.ts
```diff
@@ -549,7 +561,11 @@ export class Node {
   }
 
   getPortOfTrainrunSection(trainrunSectionId: number): Port {
-    return this.ports.find((port) => port.getTrainrunSectionId() === trainrunSectionId);
+    const port = this.ports.find((port) => port.getTrainrunSectionId() === trainrunSectionId);
+    if (port === undefined) {
+      throw new Error(`Port with trainrunSectionId ${trainrunSectionId} not found on node ${this.id}`);
```
emersion, line 572: A lot of `getPortOfTrainrunSection()` callers check the return value explicitly for `undefined`. I _think_ callers are not always guaranteed to pass a section ID which crosses the node?

To retain current code behavior, one would make the function return `| undefined` and then use `!` in all callers which don't explicitly check for `undefined`. Of course, that's not always the best course of action (e.g. if we're confident that current code will not trip on the error, or if we choose to refactor callers, or if most callers don't check for `undefined`).

src/app/models/node.model.ts
```diff
@@ -782,9 +798,15 @@ export class Node {
...
 
   getTrainrunSection(trainrun: Trainrun): TrainrunSection {
-    return this.ports
-      .find((port) => port.getTrainrunSection().getTrainrunId() === trainrun.getId())
-      .getTrainrunSection();
+    const port = this.ports.find(
+      (port) => port.getTrainrunSection().getTrainrunId() === trainrun.getId(),
+    );
+    if (port === undefined) {
+      throw new Error(
```
emersion, line 824: Contrary to `getPortOfTrainrunSection()`, the only caller of this function doesn't check for `undefined`, so it seems safe to throw an error here.

emersion (commented):
Thanks a lot for working on this! Here are a few comments.

It seems like this overlaps a bit with https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/pull/805, but that PR is blocked due to lack of time.

PR #711, opened 2025-12-31
chore: add dependabot configuration file

emersion:
Up until now, dependabot only opened pull requests for security issues. This configuration file will make it so dependabot performs regular upgrades too.


aiAdrian (approved):
Looks good!

PR #714, opened 2025-12-31
refactor: introduce `DirectedTrainrunSectionProxy` and unify time propagation

emersion:
_See individual commits._

I'm especially interested in feedback regarding the new `DirectedTrainrunSectionProxy` object. Yay/nay?

Preparation for https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/pull/589

### DIFF
--- a/src/app/services/data/trainrunsection.service.ts
+++ b/src/app/services/data/trainrunsection.service.ts
@@ -19,7 +19,7 @@ import {LogService} from "../../logger/log.service";
 import {Transition} from "../../models/transition.model";
 import {takeUntil} from "rxjs/operators";
 import {FilterService} from "../ui/filter.service";
-import {TrainrunSectionNodePair} from "../util/trainrun.iterator";
+import {DirectedTrainrunSectionProxy} from "../util/trainrun.iterator";
 import {Operation, OperationType, TrainrunOperation} from "../../models/operation.model";
 
 interface DepartureAndArrivalTimes {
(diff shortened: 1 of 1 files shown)


src/app/services/data/trainrunsection.service.ts
```diff
@@ -19,7 +19,7 @@ import {LogService} from "../../logger/log.service";
 import {Transition} from "../../models/transition.model";
 import {takeUntil} from "rxjs/operators";
 import {FilterService} from "../ui/filter.service";
-import {TrainrunSectionNodePair} from "../util/trainrun.iterator";
+import {TrainrunSectionNodePair, DirectedTrainrunSectionProxy} from "../util/trainrun.iterator";
```
louisgreiner:
```suggestion
import {DirectedTrainrunSectionProxy} from "../util/trainrun.iterator";
```

`TrainrunSectionNodePair` is unused.

louisgreiner (approved):
Considerations:
- we definitely need this kind of object to deal with source/target, left/right, forward/backward concepts without having to deal with all the combinaisons; we've actually reached the limit of the current paradigm I think
- I tested this PR, the time propagations work the same as before
- setting this new proxy in the iterator service is a good idea (instead of adding more and more things to the trainrun sections service i.e.)
- "tail" and "head" work really fine

This is really good and ok for me, thanks for the documentation and the commit messages, it makes it easy to follow 👍

src/app/services/util/trainrun.iterator.ts
```diff
@@ -13,6 +13,149 @@ export class TrainrunSectionNodePair {
...
+    return this.direction === "sourceToTarget"
+      ? this.trainrunSection.getSourceArrivalLock()
+      : this.trainrunSection.getTargetArrivalLock();
+  }
+
+  getHeadDepartureLock(): boolean {
+    return this.direction === "sourceToTarget"
+      ? this.trainrunSection.getTargetDepartureLock()
+      : this.trainrunSection.getSourceDepartureLock();
+  }
```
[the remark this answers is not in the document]
louisgreiner: Yep, very good idea

emersion:
> Proxy sounds very evil to me, could be DirectedTrainrunSection enough to refer to this actual proxy?

Why does it sound evil? The word "proxy" is widespread in the programming world.

`DirectedTrainrunSection` makes it sound like it is a kind of trainrun section, but it's actually a layer on top.

I originally went for "view" but it seemed too overloaded in an Angular codebase.

emersion:
Note, https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/pull/740 re-uses the tail/head wording but cannot leverage `DirectedTrainrunSectionProxy` because (1) the current logic operates on a `LeftAndRightTimeStructure` and changing that would require a large refactoring, and (2) the current logic operates on multiple trainrun sections.

src/app/services/data/trainrunsection.service.ts
emersion: Indeed, fixed!

src/app/services/util/trainrun.iterator.ts
```diff
@@ -2,13 +2,188 @@ import {Node} from "../../models/node.model";
...
+ * pairs when calling next():
+ *
+ *     (n₂, ts₁)
+ *     (n₃, ts₂)
+ *
+ * Iterating right-to-left starting from (n₃, ts₂) will yield the following
+ * pairs when calling next():
+ *
+ *     (n₂, ts₂)
+ *     (n₁, ts₁)
```
louisgreiner: Incredible, thanks.

louisgreiner:
LGTM! ✅

aiAdrian (approved):

PR #715, opened 2026-01-02
chore: upgrade Node to v22

emersion:
This is the latest LTS version that dependabot supports. This version is officially supported for maintenance until April 2027.

While at it, drop the patch release number so that we always get the latest bug fixes.


emersion:
Hm, this ended up not fixing dependabot PRs after all. But it's still worthwhile to upgrade.

PR #740, opened 2026-01-09
refactor: unify `onNode{Left,Right}{Departure,Arrival}TimeChanged()`

emersion:
_See individual commits._

Preliminary work for https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/pull/589

The logic of these two function pairs is exactly the same, except left
and right are swapped.

Instead of duplicating the code, introduce a common function which
operates on abstract tail and head nodes (which can be either left
or right). When onNodeLeftDepartureTimeChanged() is called, left is
tail and right is head. When onNodeRightDepartureTimeChanged() is
called, left is head and right is tail.

The common function is based on onNodeLeftDepartureTimeChanged(),
with mechanical replacements: tail replaces left, head replaces
right. No other change.

A new object LeftAndRightStructureKeys containing the mapping is
passed to the function and used when accessing fields.

This allows us to de-duplicate the logic but isn't ideal: we now
have a two-level mapping, source/target to left/right to head/tail.
We could avoid this two-level mapping by using
DirectedTrainrunSectionProxy, but this will be a much larger
refactoring. Additionally, TrainrunSectionTimesService operates
on multiple trainrun sections.

### DIFF
--- a/src/app/services/data/trainrun-section-times.service.ts
+++ b/src/app/services/data/trainrun-section-times.service.ts
@@ -173,6 +173,35 @@ export class TrainrunSectionTimesService {
     this.applyOffsetAndTransformTimeStructure();
   }
 
+  private onNodeTailArrivalTimeChanged(keys: LeftAndRightStructureKeys) {
+    this.showWarningTwoLocks = false;
+    this.roundAllTimes();
+    this.removeOffsetAndBackTransformTimeStructure();
+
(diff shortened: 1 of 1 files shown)


emersion:
Yet another option would could go for is to introduce a thin wrapper class with getters/setters similar to https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/pull/714.

kmer2016 (approved):
LGTM

PR #759, opened 2026-01-16
feat: implement and integrate new port ordering strategy

### LINKED ISSUE #635: Automatically re-order trainruns in nodes (crossings reduction)
# Description

# Mock-ups

# Acceptance Criteria

- [ ] Criteria 1
- [ ] Criteria 2
- [ ] ...

# Implementation Plan

- Step 1
- Step 2
- ...

# Tests

# Definition of Ready

- **PO/UX-UI**
  - [ ] Mock-ups are complete and validated
  - [ ] ACs are clear and have been reviewed by another refiner

- **Technical**
  - [ ] Implementation plan has been written and validated by another maintainer

- **General**
  - [ ] Validated by Adrian

jacomyal:
# Description

This PR implements a new port ordering strategy and integrates it into the UI.

The algorithm works as follows:

1. It starts from a highly connected node (highest neighbor count, then highest port count), unfolds the trainruns within it as much as possible, and propagates along train sections to minimize crossings within nodes.

2. It then looks for the largest crossings between groups of trainruns. For each crossing, it attempts to unfold it, and reapplies step 1. If this reduces the global crossing count, it saves that solution as the best.

3. It continues searching for better solutions as long as crossings remain, or until it reaches a limit on candidates tested.

The results are not as good as [LOOM](https://loom.cs.uni-freiburg.de/#stuttgart), but the algorithm is efficient and produces decent results quickly.

# Issues

- Close #635

# Checklist

### DIFF
--- a/documentation/technical/CROSSING_AWARE_ORDERING.md
+++ b/documentation/technical/CROSSING_AWARE_ORDERING.md
@@ -1,0 +1,85 @@
+# Crossing-aware port ordering
+
+This document describes the **Crossing aware** port ordering algorithm. For context on ports, their alignment, and the
+available ordering modes, see the [Ports sorting](DATA_MODEL.md#ports-sorting) section in DATA_MODEL.md.
+
+## Overview
+
+The algorithm is a greedy iterative optimizer. It first focuses on minimizing the crossings within each node, and then
+looks for alternative solutions that reduce the global amount of crossings between the nodes.
+
+It is a lightweight heuristic, not a full layout optimizer. It only reorders ports on fixed nodes with fixed positions,
+and runs fast enough to be applied on every layout update. For a more comprehensive approach to transit map layout
+(edge routing, station placement, global crossing minimization), see [LOOM](https://loom.cs.uni-freiburg.de/#stuttgart).
+
+The algorithm works in three phases:
+
+### 1. Connected components
+
+The network is split into independent connected components (via DFS). Each component is optimized separately.
+
+### 2. Port reordering
+
+For each component, a BFS traversal visits every node and reorders its ports. For each pair of ports on the same side,
+the first distinguishing criterion wins:
+
+1. Opposite transition node alignment within the node (geometric preference)
+2. Port position on the opposite side of the transition, if that side has already been ordered (with elbow correction
+   for sides that have reversed index ordering)
+3. Opposite node position (left-to-right or top-to-bottom)
+4. Port position in opposite node (if already ordered by BFS)
+5. Trainrun ordering score (tie-breaker, varies between iterations)
+
+Criteria 1 and 2 are designed to eliminate node-internal crossings by sorting ports according to the clockwise rotation
+of their transition nodes. This is what minimizes the number of crossings within each node.
+
+Criteria 3 and 4 propagate constraints along the BFS to minimize crossings between nodes. Criterion 5 is a global
+tie-breaker injected by the iterative optimizer (see below).
+
+### 3. Iterative optimization
+
+The port reordering step (phase 2) is deterministic for a given trainrun ordering. The optimizer explores different
+trainrun orderings to find one that produces fewer crossings between the nodes. It runs multiple iterations (up to
+`maxRuns`, default 50):
+
+1. Start with initial trainrun ordering (from node transitions)
+2. Reorder all ports using the current trainrun ordering as tie-breaker
+3. Count resulting crossings
+4. If crossings improved, identify largest crossing groups (contiguous sets of trainruns that cross each other), and
+   generate new candidate orderings by permuting those groups
+5. Pick the next candidate from the stack and repeat
+6. Return the configuration with the fewest crossings
+
+This is a heuristic. It uses DFS (stack-based candidate selection) and caps the number of new candidates per step
+(`maxNewCandidates`, default 10). It does not guarantee a global optimum.
+
+## Crossing detection
+
+Three types of crossings are detected:
+
+- **Direct**: Two parallel sections share the same pair of nodes, but their port order is inverted on one side.
+- **Indirect**: Two sections share one node but go to different destinations, and their port order is inverted.
+- **Node-internal**: Two transitions cross inside a node (detected via clockwise rotation check).
+
+Crossings between trainrun sections that do not share any node are not detected. For instance, a section going from top
+to bottom that visually crosses another one going from left to right will not be counted. Solving these crossings would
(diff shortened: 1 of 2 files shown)


shenriotpro:
> Before/after examples would be appreciated

I think a small benchmark with a few typical examples and the crossings count would make a lot of sense, both for users and developers

src/app/services/util/port-ordering.algo.spec.ts
```diff
@@ -0,0 +1,160 @@
...
+        },
+        trainruns: [
+          ["A1", "B1", "C1"],
+          ["A2", "B2", "C2"],
+        ],
+      });
+
+      optimizePorts(nodesArray);
+
+      // Both subgraphs processed: each B node has 1 transition, 0 crossings
```
shenriotpro: what would be the result without optimizing?

src/app/services/util/port-ordering.algo.ts
```diff
@@ -0,0 +1,342 @@
...
+ */
+export function reorderComponentPorts(
+  nodes: Node[],
+  trainrunsScores: Record<number, number> = {},
+): void {
+  const nodesWithPorts = nodes.filter((n) => n.getPorts().length > 0);
+  if (nodesWithPorts.length === 0) return;
+
+  const nodeMap = new Map(nodes.map((n) => [n.getId(), n]));
+  const sortedNodes = nodes.toSorted((a, b) => {
```
shenriotpro: It looks like we'd only need the max for each connected component, but that doesn't matter much

src/app/services/util/port-ordering.test-helpers.ts
shenriotpro: awesome

src/app/services/util/port-ordering.algo.ts
```diff
@@ -0,0 +1,342 @@
...
+type OptimizeComponentPortsOptions = {
+  maxRuns: number;
+  maxNewCandidates: number;
+};
+const DEFAULT_OPTIMIZE_COMPONENT_PORTS_OPTIONS: OptimizeComponentPortsOptions = {
+  maxRuns: 50,
+  maxNewCandidates: 10,
+};
+
+/**
```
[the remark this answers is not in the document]
jacomyal: I can, sure, but out of curiosity, why, though? It's directly and only related to the `optimizeComponentPorts` function that's defined right after. To me, it makes more sense here than in the top of the file.

src/app/services/util/port-ordering.components.ts
```diff
@@ -0,0 +1,44 @@
...
+  const components: Node[][] = [];
+
+  for (const startNode of nodes) {
+    if (visited.has(startNode.getId())) continue;
+
+    const componentNodes: Node[] = [];
+    const queue: number[] = [startNode.getId()];
+
+    while (queue.length > 0) {
+      const nodeId = queue.shift()!;
```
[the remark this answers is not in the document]
jacomyal: About efficiency, as for the other function, it searches each node once and only once, I don't see how that would be an issue. It also iterates over each node's neighbors list to find the unvisited ones, but I would have the same loop with a DFS, wouldn't I? Or am I missing something?

src/app/services/util/port-ordering.crossings.ts
```diff
@@ -0,0 +1,295 @@
...
+  PortAlignment.Bottom,
+  PortAlignment.Left,
+];
+export const ALIGNMENTS_CLOCKWISE_SCORES = new Map(
+  ALIGNMENTS_CLOCKWISE_ORDER.map((alignment, index) => [alignment, index]),
+);
+export const SWAPPED_ALIGNMENTS = new Set([PortAlignment.Bottom, PortAlignment.Left]);
+
+/**
+ * This function returns true if an alignment describes ports that are aligned horizontally. Weirdly
```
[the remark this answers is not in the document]
jacomyal: This issue is the same as for the `isElbowSwapped` function. I'll refactor the code in a way that puts both `isElbowSwapped` and `SWAPPED_ALIGNMENTS` together, with some actual clear diagram to explain the intent.

src/app/services/util/port-ordering.algo.ts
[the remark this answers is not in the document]
jacomyal: OK, it makes sense. I'll swap the order of the declaration of the functions in the file, from the most generic to the most specific, instead of the existing order that was mainly here to guarantee that a called function has already been declared (not an actual technical constraint).

jacomyal:
OK, I just pushed some additional changes, to better fit to the initial design doc, and take into account @aiAdrian's comments:
- Rename `PortOrderingType` to `OrderingAlgorithm`
- Move `OrderingAlgorithm` persistency from local storage to `NetzgrafikDto.MetadataDto`
- Move the UI control from the Settings panel to the Edit panel under `"Netzgrafik - Node optimization"`

src/app/services/data/node.service.ts
```diff
@@ -29,7 +29,8 @@ import {NODE_POSITION_BASIC_RASTER} from "../../view/rastering/definitions";
 import {MathUtils} from "../../utils/math";
 import {LabelService} from "./label.service";
 import {FilterService} from "../ui/filter.service";
-import {ConnectionDto} from "../../data-structures/technical.data.structures";
+import {ConnectionDto, OrderingAlgorithm} from "../../data-structures/technical.data.structures";
+import {optimizePorts as reorderAllPortsGlobal} from "../util/port-ordering.algo";
```
louisgreiner:
```suggestion
import {optimizePorts} from "../util/port-ordering.algo";
```

Is there a reason why this is renamed? I see in previous commit that it was former using the name as another method, but not anymore

src/app/services/util/port-ordering.algo.ts
```diff
@@ -0,0 +1,307 @@
...
+}
+
+/**
+ * This function orders all ports across all nodes using BFS traversal from a root node. To see
+ * exactly how ports are ordered in a single node (where the logic actually is), check
+ * reorderNodePorts.
+ */
+function reorderComponentPorts(nodes: Node[], trainrunsScores: Record<number, number> = {}): void {
+  const nodesWithPorts = nodes.filter((n) => n.getPorts().length > 0);
+  if (nodesWithPorts.length === 0) return;
```
SharglutDev: Can this condition happen ? If I understood correctly, `getConnectedComponents` returns all connected nodes so they should have at least one port, shouldn't they ?

If that's the case, maybe we could return an error somewhere or add a test which should fail if we get a node with no ports in the array.

PR #805, opened 2026-02-04
refactor: fix null type errors

Synar:
Properly handle null and undefined values in many functions without lying on the types, and without ignoring the possibility of such values which can lead to crashes (such as https://github.com/OpenRailAssociation/osrd/issues/14085 which this pr fixes)

There are a few arbitrary decisions taken in this pr which can be debated. In particular, how much we want to throw or use "!" VS how much we want to fallback or use "?". I generally tried to assume that getters could be called before init and so avoided as much as possible to fail on null values for these, but for a few setters I was more lax. The fallback values when sorting are a bit ugly too as a workaround.

The pr can be checked by setting ` "strictNullChecks": true` in tsconfig.json. All edited files should have 0 errors after this pr, except node.service which is only partially handled by this pr.

### DIFF
--- a/src/app/data-structures/business.data.structures.ts
+++ b/src/app/data-structures/business.data.structures.ts
@@ -178,11 +178,11 @@ export interface NodeDto {
   transitions: TransitionDto[]; // all tranisitons aligned to the node
   connections: ConnectionDto[]; // all connections aligned to the node
 
-  resourceId: number; // reference to the algined (resource - not yet implemented)
+  resourceId: number | null; // reference to the algined (resource - not yet implemented)
   perronkanten: number; // number of tracks where train can stop
   connectionTime: number; // aka Umsteigezeit - time used to change train in minutes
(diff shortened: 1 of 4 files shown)


louisgreiner (approved):
LGTM, thanks

The commit messages need a slight overwork, we use Conventional Commit, then this should be "refactor", not "refacto":
https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/blob/999e8198c32ddb460532be150a20106d9d0afcbe/commitlint.config.ts#L10

emersion (commented):
Nice work on the `sortPorts()` cleanup!

Can the cleaned up files be added to `tsconfig.strictNullChecks.json`, to ensure they don't regress?

Pivouane (approved):
LGTM ! 
Be careful as it is very likely that this PR needs to be rebased

aiAdrian:
If this is still required - please update/comment otherwise we might close it, because is one more than 5month @Synar .

PR #828, opened 2026-02-19
827 overuse of atsource boolean var and more

### LINKED ISSUE #827: Overuse of "atSource" boolean var (and more)
At some places in the code, we use `atSource` (boolean) to deal with source/target expected behaviors. However, this can lead to misunderstanding when the return type of a method should return a boolean as well.

For example:
https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/blob/52e468b2149e31592cd8d2787f227a92ef248de9/src/app/view/editor-main-view/data-views/trainrunsections.view.ts#L2022-L2026C11
```ts
  private filterTimeTrainrunsectionNonStop(
    viewObject: TrainrunSectionViewObject,
    atSource: boolean,
    isArrival: boolean,
  ): boolean {
```

I don't know what should be done

aiAdrian:
This refactoring replaces the ambiguous `atSource: boolean` parameter with the explicit
`TrainrunSectionNodeAnchor` enum. This improves readability, reduces boolean ambiguity,
and makes the intent of each method call clearer. Several methods in
`TrainrunSectionsView`, `PerlenketteSectionComponent`, and related view logic were updated
accordingly to use the new enum-based API.

### DIFF
--- a/src/app/data-structures/technical.data.structures.ts
+++ b/src/app/data-structures/technical.data.structures.ts
@@ -27,6 +27,14 @@ export enum TrainrunSectionText {
 }
 
 /**
+ * Classifies which anchor (node) of a TrainrunSection is being evaluated.
+ */
+export enum TrainrunSectionNodeAnchor {
+  Source = "source",
+  Target = "target",
+}
+
+/**
  * Represents a mapping of TrainrunSectionText and PointDTO used in the "cached" path : PathDto.
  * The PathDto is used in the TrainrunSectionDto. The application (editor) recalculates after each
  * position update the path and stores the positions of the text in this mapping.
--- a/src/app/view/editor-main-view/data-views/trainrunsections.view.ts
+++ b/src/app/view/editor-main-view/data-views/trainrunsections.view.ts
@@ -13,6 +13,7 @@ import {Vec2D} from "../../../utils/vec2D";
 import {
   ColorRefType,
   PortAlignment,
+  TrainrunSectionNodeAnchor,
   TrainrunSectionText,
 } from "../../../data-structures/technical.data.structures";
 import {StaticDomTags} from "./static.dom.tags";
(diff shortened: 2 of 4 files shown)


louisgreiner (commented):
After a very very short review, I think replacing the true/false values by an actual Type increases the readability, however the main issue remains. I'm not sure this is how we should treat that. Also, I don't really know what would be better 😄 

@emersion @SharglutDev might have better insights, even tho the issue is more at a conceptual level

src/app/view/editor-main-view/data-views/trainrunsections.view.ts
```diff
@@ -1206,24 +1242,29 @@ export class TrainrunSectionsView {
...
           ")",
       )
       .attr(StaticDomTags.EDGE_NODE_ID, (d: TrainrunSectionViewObject) =>
-        TrainrunSectionsView.getNode(d.trainrunSection, atSource).getId(),
+        TrainrunSectionsView.getNode(d.trainrunSection, anchorNode).getId(),
+      )
+      .classed(
+        StaticDomTags.EDGE_IS_TARGET,
+        anchorNode === TrainrunSectionNodeAnchor.Source
+          ? TrainrunSectionNodeAnchor.Target
```
louisgreiner: This does not return a boolean value, which is expected for setting the value of `StaticDomTags.EDGE_IS_TARGET`

louisgreiner (commented):
(I don't have to time to fully review)

The test files don't build and some typing issue lies in this PR (which are not detected since we do not have enough type checking in the CI).

I'm a bit worried about making these changes that affect so many things so quickly, I would recommend establish a clear strategy, and maybe split in smaller commit and make sure all the commits compile. This would ensure that we don't break too many things

louisgreiner (commented):
Except these two nits, this looks good to me. Maybe we can point all the places where we should use this, and list them in an refactor issue

src/app/data-structures/technical.data.structures.ts
```diff
@@ -26,6 +26,14 @@ export enum TrainrunSectionText {
   TrainrunSectionNumberOfStops,
 }
 
+/**
+ * Classifies which anchor (node) of a TrainrunSection is being evaluated.
+ */
+export enum TrainrunSectionNodeAnchor {
```
emersion, line 32: Using an enum here means that function calls are quite a bit longer, e.g.

```ts
const v1 = TrainrunSectionsView.enforceStartTextAnchor(ts, TrainrunSectionNodeAnchor.Target);
```

If we wanted to keep these a bit shorter, we could use a union instead:

```ts
export type TrainrunSectionAnchor = 'source' | 'target';
```

Function calls would look like this:

```ts
const v1 = TrainrunSectionsView.enforceStartTextAnchor(ts, 'target');
```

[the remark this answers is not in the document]
emersion: `TrainrunSectionEndpoint` would be another possible name. `TrainrunSectionAnchor` looks fine to me as well.

emersion (commented):
Overall this approach looks good to me, very nice to get rid of these boolean flags!

aiAdrian:
@louisgreiner @emersion -> what is the next step we should do to close this issue? Any idea / support are welcome

louisgreiner:
@aiAdrian, you have some review comments that are not resolved I guess, if you need help for dealing with them, I can help.

aiAdrian:
This fix can not be merged - because we have to make a bigger refactoring first. Then this will might be no longer relevant.

PR #858, opened 2026-02-25
fix: Error after importing a reticular similar to an existing one and…

### LINKED ISSUE #851: Error after importing a reticular similar to an existing one and clicking on a node.
### What is the input mode?

Touchpad

### What type of issue is it?

Functionality

### What happened?

When you export the existing reticular and then re-import it, if you click on a node, an invalid train that starts and ends at this node is created. This invalid train is not saved if you refresh the page.
It seems to be related to cache management.

NB: This bug was first triggered in OSRD when the user changed the language.

Nothing should happen.

1. Export the reticular
2. Re-import it
3. Click on a node

Local

Firefox

2.10.15

aiAdrian:
… clicking on a node fixed

### Analysis

The error occurred sporadically, and until now no user was able to report it in a reproducible way.  
I have now understood and identified the root cause.
 
 

### Issue is here
https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/blob/a65706ef5b296e4eb36306b60b41d4f9f2288551/src/app/view/editor-main-view/data-views/nodes.view.ts#L731

### check is wrong (unsafe)

There is a startNode not eq undefined thus the check is not true when an object copy will be returned (start or end node) 

```typescript 
If (startNode === endNode) {
```

### check must be (safe)

The ID check itself is fine. But to ensure that the objects are valid we have also to check that null or undefined values are handled correctly  

```typescript
if (!startNode || !endNode || startNode?.getId() === endNode?.getId()) {
```

### fix 

to ensure that also the check with the source / target node can not cause another issue we fix it as well

```typescript
    if (startNode?.getId() === endNode?.getId()) {
      return;
    }
    if (existingTrainrunSection !== null) {
      if (
        existingTrainrunSection.getSourceNodeId() === startNode?.getId() &&
        existingTrainrunSection.getTargetNodeId() === endNode?.getId()
      ) {
        return;
      }
      if (
        existingTrainrunSection.getSourceNodeId() === endNode?.getId() &&
        existingTrainrunSection.getTargetNodeId() === startNode?.getId()
      ) {
        return;
      }
    }
```

### DIFF
--- a/src/app/services/data/data.service.ts
+++ b/src/app/services/data/data.service.ts
@@ -151,13 +151,7 @@ export class DataService implements OnDestroy {
     this.trainrunSectionService.initializeTrainrunSectionRouting();
     this.nodeService.validateAllConnections();
     this.trainrunService.propagateInitialConsecutiveTimes();
-    this.nodeService.nodesUpdated();
-    this.nodeService.transitionsUpdated();
-    this.nodeService.connectionsUpdated();
-    this.trainrunSectionService.trainrunSectionsUpdated();
-    this.noteService.notesUpdated();
-    this.labelService.labelUpdated();
-    this.labelGroupService.labelGroupUpdated();
+    this.triggerViewUpdate();
 
     this.netzgrafikColoringService.generateGlobalStyles(
       this.getTrainrunCategories(),
--- a/src/app/services/data/data.service.ts
+++ b/src/app/services/data/data.service.ts
@@ -40,6 +40,10 @@ export class DataService implements OnDestroy {
 
   private destroyed = new Subject<void>();
 
+  // Counter to force D3 data rebinding when loading new DTO (not for undo/redo)
+  // Fixes issue #851: prevents stale object references in D3 callbacks
+  private netzgrafikLoadCounter = 0;
+
   private readonly netzgrafikLoadedInfoSubject = new BehaviorSubject<NetzgrafikLoadedInfo>(
     new NetzgrafikLoadedInfo(true, false),
   );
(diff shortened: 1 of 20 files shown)


emersion:
This PR seems to workaround the problem, rather than fixing the root cause. There are stale cached objects held by d3, and these aren't properly nuked when loading a new DTO.

louisgreiner (commented):
I'm not getting the matter of the issue yet, but I'll keep an eye on it.

aiAdrian:
> This PR seems to workaround the problem, rather than fixing the root cause. There are stale cached objects held by d3, and these aren't properly nuked when loading a new DTO.
 
```typescript
      .on("mouseover", (n: NodeViewObject) => this.onHoverNodeMouseover(n.node))
      .on("mouseout", (n: NodeViewObject) => this.onHoverNodeMouseout(n.node))
      .on("mousedown", (n: NodeViewObject) => this.onNodeMousedown(n.node))
      .on("mouseup", (n: NodeViewObject) => this.onNodeMouseup(n.node))
      .on("dblclick", (n: NodeViewObject) => this.onNodeDetailsClicked(n.node));
```

src/app/services/data/data.service.ts
```diff
@@ -151,13 +151,7 @@ export class DataService implements OnDestroy {
...
     this.nodeService.validateAllConnections();
     this.trainrunService.propagateInitialConsecutiveTimes();
-    this.nodeService.nodesUpdated();
-    this.nodeService.transitionsUpdated();
-    this.nodeService.connectionsUpdated();
-    this.trainrunSectionService.trainrunSectionsUpdated();
-    this.noteService.notesUpdated();
-    this.labelService.labelUpdated();
-    this.labelGroupService.labelGroupUpdated();
+    this.triggerViewUpdate();
```
aiAdrian: It found by hazard this duplicated code. While testing to "clear" all object during load which could overcome the problem. But this leads to more complicated things to do. This was the reason i stop changed the loading stuff. 

I think we have just to ensure when then callbacks gets correctly handled. Once we face a comparison  by object in the code where we compare "rendering system stored / cached objects" with ...Service stored we should focus on id based comparision not on objects. Thus we haven't to change the loading stuff. This problem was located at wrong comparision logic.

aiAdrian:
@emersion @louisgreiner  This Problem is still open and should be fixed - the problem is we have to find a better variant. We might can make a short call next week to discuss. If there are others interested -> please write into a comment.

emersion:
Alternative fix: https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/pull/1009

louisgreiner:
Closed in favor of https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/pull/1009

PR #862, opened 2026-02-26
fix: ensure both trainruns have endNodes at "pin"

aiAdrian:


emersion:
How to test this bugfix?

aiAdrian:
I have to make a demo -> call

PR #890, opened 2026-03-18
fix: check missing transitions after deletion test

Math-R:
Fix the failing check missing transitions after deletion test


louisgreiner (commented):
With that change, we are back to the initial issue:

https://github.com/user-attachments/assets/9521f8b3-f42f-48a6-8f17-7200297c2996

PR #903, opened 2026-03-26
refactor: replace analytics shortest path algorithms with OD graph algorithms

### LINKED ISSUE #784: Refactoring Shortest travel times journey chain using OD Matrix algorithm
# Description

> [!NOTE]
> The "Shortest travel times journey chain" is also refered as "Analytics" feature (which is actually the name used in the application).

The current "Analytics" feature:
- uses a separate, slower, and unreliable algorithm,
- produces different results than the OD Matrix,
- loses its rendered results when the user pans/zooms the network - the rendering integration is poor - just quick & dirty implemented.

The "Analytics" feature should be refactored to display to travel times and the connections given by the OD Matrix algorithm. Unify the “Analytics” shortest‑path feature with the OD Matrix algorithm so both features use the same computation logic, produce identical travel‑time results, and behave consistently in the UI. The current code should be totally replaced by this, to use a single algorithm for the both features.

The next version of Netzgrafik "Analytics" will only support interactions based on clicking a node.

The following screenshots have been made using this network: [analytics.json](https://github.com/user-attachments/files/24885389/analytics.json). When clicking on the "Analytics" button, the view changes, highlighting the departure minutes:

Image

Then we can click on any node (here on node A), and this shows the shortest path to other nodes. Using this shortest path, (1) represents the number of connections to this node and (2) represents the travel time to this node.

Image

However, when we look at the Origin-Destination Matrix for the same network, the results are not the same. For instance, if we click on node F:

Image

The OD Matrix does not output the same result (⚠️ make sure to click on "Temps de parcours", to observe only the travel times, and not the "cost", which is `travel time + connections * penality`)

Image

As we have a better trust on the OD Matrix algorithm, we need both features to use the same algorithm, then displaying the same results. The goal is then to refactor the OD Matrix algorithm, to be able to use it elsewhere and generate the shortest path from the clicked node to all other nodes (and not from all nodes to all nodes, which is currently what the OD Matrix algorithm doing).

# Mock-ups

The rendering should not be affected by this refactoring, except that the color of the nodes after computation should be based on the travel time, using the same color scale as the OD Matrix does.

# Acceptance Criteria

- [ ] The OD Matrix feature works the same and display the same results as before the refactoring
- [ ] The "Analytics" feature works the same as before and display the same results as the OD Matrix feature, ***without taking connection penality into account***
  - [ ] Clicking on a node compute "*one‑to‑all* shortest paths computation"
    - [ ] In the nodes, the "Total travel time" and "Connections" are displayed:
      - [ ] "Connections" (number of transfers): displayed on the left side of the node as `#n`
      - [ ] "Total travel time" (travel time without any penalty): displayed on the right side of the node, formatted as `hh:mm`.
    - [ ] The nodes then are colored, using the same color scale as the OD Matrix does, based on the "Total travel time" value
  - [ ] The trainrun sections part of the computed shortest paths are highlighted (as it is already the case with old "Analytics" feature)
  - [ ] The transitions part of the computed shortest paths are highlighted (as it is already the case with old "Analytics" feature)
- [ ] "Analytics" feature should take active filters into account (as it is already the case for OD Matrix feature)
- [ ] Once the "Analytics" is activated and a result is there, it remains visible until another node is clicked, or "Analytics" is explicitly deactivated
- [ ] "Analytics" only supports interactions based on clicking a node
- [ ] Paning and zooming don't affect the results rendering
- [ ] Exports (`.png` and `.svg`) allow rendered results to be exported

# Implementation Plan

- Refactor the OD Matrix algorithm service to make it possible to generate the shortest path from a node to all other nodes (and not from all nodes to all other nodes) and make it available from a service.
  - The OD Matrix service should act like an interface that offers:
    - *all‑to‑all* shortest‑path computation (existing behavior)
    - *one‑to‑all* computation (a single row of the matrix)
    - ⚠️ the core algoritm should not be modified, just the endpoints
  - [`src/app/services/analytics/origin-destination/components/origin-destination.service.ts`](https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/blob/main/src/app/services/analytics/origin-destination/components/origin-destination.service.ts)
- Remove all the existing "Analytics" algorithm code, and use the above service instead
  - [`src/app/services/analytics/analytics.service.ts`](https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/blob/main/src/app/services/analytics/analytics.service.ts)
- Populate the "Analytics" rendered values with the OD Matrix algorithm service
- Update the tests (or create new ones) to test the unified algorithm and both endpoints

# Tests

# Definition of Ready

- **PO/UX-UI**
  - [ ] Mock-ups are complete and validated
  - [ ] ACs are clear and have been reviewed by another refiner

- **Technical**
  - [ ] Implementation plan has been written and validated by another maintainer

- **General**
  - [ ] Validated by Adrian

# Outscope TODOs

This will need to create another issue:
- Extended UI
  - such as checkboxes or switches, for the "value" fields and "color" parameter
- Prepare Analytics component for future extension (difficult to imagine from that)

jacomyal:
Close https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/issues/784

This PR refactors the analytics shortest-path computation, to reuse the existing OD graph infrastructure (`buildEdges`, `computeNeighbors`, `topoSort`, `computeShortestPaths`) instead of the standalone `ShortestTravelTimeSearch` algorithm.

- The first commit adds path tracking to the OD graph's shortest-path output, so that it can report which trainrun sections are traversed
- The second commit replaces the analytics service internals with a pipeline over these shared utilities
- The third commit deletes the no-longer used shortest-paths algorithm files

---

#### 1. More parallel highlighted sections

In some `A->B->C` cases, the OD graph algorithm sometimes can _rightfully_ find a path from A to C that does not follow the shortest path from A to B, because it now starts "counting the cost" at the departure time of the first trainrun section, rather than 00:00 all the time, as it did before:

nge-screenshot-1

#### 2. Some unavoidable duplications

The algorithm fails to use existing shortest subpaths as a tiebreaker for longer paths. So, in some cases, a long shortest path can highlight some sections that are parallel to a "shorter shortest path" (i.e. a shortest path to some node closer to the start node). This is a limitation that sometimes adds some noise to the final visual output. I did not find a simple enough way to mitigate it, though, but it certainly can be done later.

In the following example, both the path 1 (IC81) and path 2 (IR15) are highlighted from Bern to RTR. The path 1 makes a lot of sense, because it belongs to the actual shortest path from Lausanne (start node) to RTR. But both paths 1 and 2 can be used, in the shortest path from Lausanne to Langent., and here, the algorithm selects the path 2. And improved version could only highlight the path 1 in this very case.

nge-screenshot-2


aiAdrian (approved):
Very nice work. 1st step towards the correct implementation is done! 

- I just got, that for trainruns which start not at zero minute have a departure offset -> this means the traveltime is 0 at departure node -> dep at n -> travel time m -> arrival time n+m -> shortest travel time n+m 

-> we have to discuss this 

-> the rendering is not yet fine - but as good as it was :-) -> we have to discuss how to improve it later 

THANK YOU VERY MUCH -> nice work !

src/app/view/util/origin-destination-graph.ts
```diff
@@ -205,17 +220,51 @@ export const computeShortestPaths = (
         ) {
           connection = 1;
         }
-        dist.set(neighbor, [alt, dist.get(vertex)[1] + connection]);
+        const newConnections = dist.get(vertex)[1] + connection;
+        // We use the connections count as a tiebreaker for paths with equal
```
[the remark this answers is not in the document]
jacomyal: When I was trying to deduplicate the paths on screen (to resolve the `2. Some unavoidable duplications` described in the PR), I first tried this basic tiebreaker, which in some case I had actually resolved the issue.

I later decided to keep it, as having some tiebreaker makes the final decision more explainable, and this one in particular felt meaningful to me, while remaining quite easy to implement and maintain, basically.

src/app/view/util/origin-destination-graph.ts
```diff
@@ -148,19 +156,25 @@ export const topoSort = (graph: Map<Vertex, [Vertex, number][]>): Vertex[] => {
...
+): Map<number, [number, number, number[]]> => {
   const tsPredecessor = new Map<number, number>();
   tsSuccessor.forEach((v, k) => {
     tsPredecessor.set(v, k);
   });
-  const res = new Map<number, [number, number]>();
+  // Maps each reached nodeId to its arrival "convenience" vertex.
+  const destinations = new Map<number, Vertex>();
+  // Maps each vertex to its best predecessor, for path reconstruction.
+  const prev = new Map<Vertex, Vertex>();
```
shenriotpro, line 176: nice

src/app/view/util/origin-destination-graph.ts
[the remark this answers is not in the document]
shenriotpro: I see, thanks, I think a comment to explain the usecase would be nice

src/app/services/analytics/analytics.service.ts
```diff
@@ -28,36 +31,76 @@ export class AnalyticsService implements OnDestroy {
...
-    const shortestTravelTimeSearch = new ShortestTravelTimeSearch(
+    this.findAllShortestDistanceNodesSubject.next(
+      this.findShortestDistanceNodes(departureNodeId, trainrunSectionId),
+    );
+  }
+
+  /**
+   * Compute shortest paths from a departure node using the OD graph algorithm.
+   * When startTrainrunSectionId is set, the first hop is constrained to that
+   * section, and paths back to the departure node are excluded.
```
shenriotpro, line 49: why are we excluding those?

Math-R (approved):
Overrall looks good to me , tested it works well :)
But I'm not sure on this one I may have miss some specifities.

Maybe @emersion could be more accurate

src/app/services/analytics/analytics.service.ts
jacomyal: If we accept those, then the results will including results that go "backward", reaching nodes that are accessible only through leaving the starting node, then going back to it, then going to another direction, basically.

I chose that rule to get results kind of similar to what we had with the original analytics algorithm.

louisgreiner (commented):
Tested, works very well.

Already created a related enhancement (not to be dealt in this PR): https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/issues/925

louisgreiner:
Added a mention of https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/issues/784

louisgreiner:
Does this PR need more review? I feel like this looks quite ready to merge

PR #902, opened 2026-03-25
feat: change the traffic side 

Uriel-Sautron:
This PR allows to change the traffic side in the reticular

Closes https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/issues/771

PR #916, opened 2026-04-08
refactor: create time stepper component

### LINKED ISSUE #923: create time stepper component
### Who would benefit from this feature?

Developers

### What is this feature about?

Create a new component to avoid code duplication and facilitate the development of the traffic side feature.

### Why is this feature valuable?

Make the code easier to maintain

Uriel-Sautron:
Create a new composant and use in `trainrun-section-tab`, facilitate the development of [traffic-side feature](https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/pull/902)
close #923

### DIFF
--- a/src/app/view/dialogs/trainrun-and-section-dialog/trainrunsection-tab/time-stepper/time-stepper.component.html
+++ b/src/app/view/dialogs/trainrun-and-section-dialog/trainrunsection-tab/time-stepper/time-stepper.component.html
@@ -1,0 +1,20 @@
+<ng-container *ngIf="isDisplayed">
+  <div [class]="containerClass" class="time-stepper-container">
+    <button tabindex="-1" mode="ghost" class="step-button" (click)="decrement.emit()">-</button>
+    <input
+      [tabindex]="tabIndex"
+      [id]="inputId"
+      [ngClass]="inputClass"
+      type="number"
+      sbbInput
+      spellcheck="false"
+      pattern="^[0-9.]*"
+      [min]="min"
+      minlength="1"
+      [(ngModel)]="value"
+      (change)="onValueChanged()"
+      #inputElement
+    />
+    <button tabindex="-1" mode="ghost" class="step-button" (click)="increment.emit()">+</button>
+  </div>
+</ng-container>
--- a/src/app/view/dialogs/trainrun-and-section-dialog/trainrunsection-tab/time-stepper/time-stepper.component.ts
+++ b/src/app/view/dialogs/trainrun-and-section-dialog/trainrunsection-tab/time-stepper/time-stepper.component.ts
@@ -1,0 +1,39 @@
+import {Component, ElementRef, EventEmitter, Input, Output, ViewChild} from "@angular/core";
+
+@Component({
+  selector: "sbb-time-stepper",
+  templateUrl: "./time-stepper.component.html",
+  styleUrls: ["./time-stepper.component.scss"],
+})
+export class TimeStepperComponent {
+  @Input() isDisplayed = true;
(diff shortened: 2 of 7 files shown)


emersion:
Could you add a reference to the parent issue in the PR description?

louisgreiner (approved):
Nice! LGTM ✅

aiAdrian (approved):
LGTM

src/app/view/dialogs/trainrun-and-section-dialog/trainrunsection-tab/time-stepper/time-stepper.component.ts
```diff
@@ -0,0 +1,45 @@
...
+  @Input() inputClass: string | string[];
+  @Input() tabIndex: number;
+  @Input() type = "number";
+  @Input() pattern = "^[0-9.]*";
+  @Input() spellcheck = false;
+  @Input() min: number | string;
+  @Input() max: number | string;
+  @Input() minLength: number | string;
+  @Input() maxLength: number | string;
+  @Input() value: number;
```
emersion, line 14: It would seem to me like most of these are always set to the same value regardless of the use? I think it would make more sense to only expose an `@Input` if we need to customize the time stepper.

src/app/view/dialogs/trainrun-and-section-dialog/trainrunsection-tab/time-stepper/time-stepper.component.ts
```diff
@@ -0,0 +1,45 @@
...
+
+  @Output() valueChange = new EventEmitter<number>();
+  @Output() changed = new EventEmitter<void>();
+  @Output() decrement = new EventEmitter<void>();
+  @Output() increment = new EventEmitter<void>();
```
emersion, line 20: All of the `onXXXButtonPlus`/`Minus` callbacks in `TrainrunSectionTimesService` do exactly the same thing. I think it would make sense to drop these event emitters, and instead internally compute the new value and then trigger `changed`.

But that's a larger refactoring, so I'd be fine leaving this for another PR.

src/app/view/dialogs/trainrun-and-section-dialog/trainrunsection-tab/time-stepper/time-stepper.component.ts
Uriel-Sautron: Done

src/app/view/dialogs/trainrun-and-section-dialog/trainrunsection-tab/time-stepper/time-stepper.component.html
```diff
@@ -0,0 +1,20 @@
...
+    <button tabindex="-1" mode="ghost" class="step-button" (click)="decrement.emit()">-</button>
+    <input
+      [tabindex]="tabIndex"
+      [id]="inputId"
+      [ngClass]="inputClass"
+      type="number"
+      sbbInput
+      spellcheck="false"
+      pattern="^[0-9.]*"
+      min="0.0001"
```
emersion: Ah actually I think that one we still need to expose an `@Input` for: we want to allow a departure/arrival time of 0! (But not a travel time)

emersion (approved):
LGTM apart from that comment!

src/app/view/dialogs/trainrun-and-section-dialog/trainrunsection-tab/time-stepper/time-stepper.component.html
Uriel-Sautron: Oupss good catch, done

PR #918, opened 2026-04-08
fix(service): fix travelTimeOffset for asymmetric trainruns

aschantraine:
Fixes the calculation of `travelTimeOffset` to use `getDirectedTrainrunSectionProxy().getTravelTime()` instead of `trainrunSection.getTravelTime()` ensuring the correct travelTime is used (forward or backward).
Adds a test to check consecutive time with travel time over 60min for an asymmetric run


emersion (approved):
Very nice work, LGTM!

aiAdrian (approved):
LGTM

Issue #886, opened 2026-03-17
Upgrade Angular


emersion:
We're currently stuck at version 17. Latest version to 21.

- [x] v18 (https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/pull/924)
- [x] v19 (https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/pull/970)
- [x] v20 (https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/pull/1002)
- [x] v21 (https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/pull/1272)
- [x] v22 (https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/pull/1286)

See https://angular.dev/update-guide?v=17.0-21.0&l=3

emersion:
We've caught up all Angular upgrades now!

PR #924, opened 2026-04-10
chore: upgrade to Angular v18

emersion:
Most of the changes were performed automatically by running:

    npx ng update @angular/core@18 @angular/cli@18 @angular-eslint/schematics@18 angular-server-side-configuration@18 @sbb-esta/angular@18 @angular/cdk@18
    npm run format

See this upgrade checklist:
https://angular.dev/update-guide?v=17.0-18.0&l=3

One manual change was required for I18nService. Angular v18 now uses Vite instead of Webpack, so webpackInclude dynamic imports are not working anymore. Instead, @rollup/plugin-dynamic-import-vars needs to be used:
https://github.com/rollup/plugins/tree/master/packages/dynamic-import-vars

Only relative dynamic imports are supported, so we need to convert absolute paths. Additionally, create one tiny .js wrapper per language for the locale data to avoid manually searching through the node_modules directory.

References: https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/issues/886


louisgreiner (commented):
Thanks for the exhaustive description, it makes it easier to understand.

While going through the check-list, I wonder about this line:
> - [ ] Remove expressions that write to properties in templates that use [(ngModel)]

But we use `[(ngModel)]` still a lot. Should we do something about it?

Besides that, this looks good to me!

emersion:
> But we use [(ngModel)] still a lot. Should we do something about it?

`[(ngModel)]` is valid and continues to be supported by Angular. This list item is about some invalid uses of `[(ngModel)]` which were previously accepted but are no longer accepted. See https://github.com/angular/angular/tree/18.0.3/packages/core/schematics/migrations/invalid-two-way-bindings

The migration tool hasn't found any of these invalid uses.

louisgreiner (approved):
Thanks for the explanations! LGTM

aiAdrian (approved):
Tested and approved - Thank you nice work!

kmer2016 (approved):
LGTM

louisgreiner (approved):

emersion:
(Trivial rebase against `main`)

PR #930, opened 2026-04-14
refacto(trainrunsection): replaceintermediatestopwithnode, simplify isnonstop using optionnal boolean and rename noStopduration

Math-R:


src/app/services/data/trainrunsection.service.ts
```diff
@@ -989,9 +990,11 @@ export class TrainrunSectionService implements OnDestroy {
     const node2 = trainrunSection1.getTargetNode();
     const nodeIntermediate = this.nodeService.getNodeFromId(nodeId);
     const transition1: Transition = node1.getTransition(trainrunSection1.getId());
-    const nonStop1 = transition1 !== undefined ? transition1.getIsNonStopTransit() : false;
+    transition1?.setIsNonStopTransit(isNonStop);
+    const nonStop1 = transition1?.getIsNonStopTransit() ?? false;
     const transition2: Transition = node2.getTransition(trainrunSection1.getId());
-    const nonStop2 = transition2 !== undefined ? transition2.getIsNonStopTransit() : false;
+    transition2?.setIsNonStopTransit(isNonStop);
+    const nonStop2 = transition2?.getIsNonStopTransit() ?? false;
```
louisgreiner: Starting with this situation: `node1 --(section1)-- node2`

During the process, `transition1` refers to the transition of section1 in node1 and `transition2` refers to the transition of section1 in node2.

Then end up with this situation: `node1 --(section1)-- nodeIntermediate --(section2) -- node2`

I could not fully test, so I might be wrong, but I'm surprised that both `nonStop1` and `nonStop2` need to be adapted in order to make a stop in the intermediate node.

Math-R: Finally i revert my changes and apply the isNonStopTransit = false to the nodeintermediate transition

louisgreiner (commented):
The current implementation seems not to work, you can try on uploading this file and add an intermediate stop: [networkGraphic(27).json](https://github.com/user-attachments/files/26754771/networkGraphic.27.json)

Math-R:
> The current implementation seems not to work, you can try on uploading this file and add an intermediate stop: [networkGraphic(27).json](https://github.com/user-attachments/files/26754771/networkGraphic.27.json)

Yes sorry i womehow remove the `!` before `isNonStop` in the condition, now its fix

louisgreiner:
I did not have the time to test the [latest](https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/pull/930#issuecomment-4259932600) change, but I had that in mind: https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/commit/849ede8f23ce9dc90f6e3b24490dacdf61c87b3b

What do you think?

src/app/services/data/data.service.ts
```diff
@@ -138,7 +138,8 @@ export class DataService implements OnDestroy {
           this.trainrunSectionService.replaceIntermediateStopWithNode(
             currentSection.id,
             newNode.getId(),
-            undefined,
+            true,
```
emersion, line 141: `stopDuration` was `undefined` before, and now `noDurationStop` is `true`. Why is that?

src/app/view/editor-main-view/data-views/editor.keyEvents.ts
```diff
@@ -264,7 +264,11 @@ export class EditorKeyEvents {
 
       sections.forEach((ts) => {
         ts.setNumberOfStops(ts.getNumberOfStops() + 1);
-        this.trainrunSectionService.replaceIntermediateStopWithNode(ts.getId(), node.getId());
+        this.trainrunSectionService.replaceIntermediateStopWithNode(
+          ts.getId(),
+          node.getId(),
+          false,
```
emersion, line 270: Nit: `false` is the default value.

src/app/services/data/trainrunsection.service.ts
```diff
@@ -1116,7 +1122,7 @@ export class TrainrunSectionService implements OnDestroy {
       interpolatedPosition.getY(),
     );
 
-    this.replaceIntermediateStopWithNode(trainrunSection.getId(), newNode.getId(), 0);
+    this.replaceIntermediateStopWithNode(trainrunSection.getId(), newNode.getId(), true, false);
```
emersion, line 1125: With this last parameter set to `false`, it seems like `replaceIntermediateStopWithNode()` is never called with `isNonStop` set to `true`?

src/app/services/data/trainrunsection.service.ts
```diff
@@ -1033,8 +1038,9 @@ export class TrainrunSectionService implements OnDestroy {
     travelTime1 = travelTime1 < 0 ? travelTime2 : travelTime1;
     travelTime2 = travelTime2 < 0 ? travelTime1 : travelTime2;
     const calculatedTravelTime = Math.min(travelTime1, travelTime2);
-    const halteZeit =
-      stopDuration ?? Math.min(minHalteZeitFromNode, Math.max(0, calculatedTravelTime - 2));
+    const halteZeit = noDurationStop
+      ? 0
+      : Math.min(minHalteZeitFromNode, Math.max(0, calculatedTravelTime - 2));
```
emersion, line 1043: The original code in `main` doesn't have this special case. Does anybody remember why it got added?

If not, can we drop it?

emersion (commented):
The PR title only references a function name without more details. It would be nice to make it more explicit.

The second commit description lacks an explanation about what it fixes and why. It would be nice to record that information.

src/app/services/data/trainrunsection.service.ts
```diff
@@ -969,7 +969,8 @@ export class TrainrunSectionService implements OnDestroy {
   replaceIntermediateStopWithNode(
     trainrunSectionId: number,
     nodeId: number,
-    stopDuration?: number,
+    noDurationStop = false,
+    isNonStop = false,
```
emersion: Either these are all false, either these are all true, so we probably only need a single `noDurationStop` boolean here instead of two?

Math-R:
@emersion @louisgreiner 
I made the changes following your [recommendations](https://github.com/osrd-project/netzgrafik-editor-frontend/pull/26#discussion_r3111492563)

src/app/models/node.model.ts
```diff
@@ -574,16 +574,17 @@ export class Node {
...
     transition.setPort1Id(port1.getId());
     transition.setPort2Id(port2.getId());
-    transition.setIsNonStopTransit(
-      isNonStop
-        ? true
-        : this.trainrunCategoryHaltezeiten[trainrun.getTrainrunCategory().fachCategory].no_halt,
-    );
+    const defaultIsNonStop =
+      trainrun?.getTrainrunCategory()?.fachCategory !== undefined
+        ? (this.trainrunCategoryHaltezeiten[trainrun.getTrainrunCategory().fachCategory]?.no_halt ??
```
emersion: Why this change? Wasn't `this.trainrunCategoryHaltezeiten[trainrun.getTrainrunCategory().fachCategory].no_halt` working fine before?

`trainrun` cannot be null, and I don't think `trainrun.getTrainrunCategory()` could be null either.

emersion (commented):
LGTM apart from this minor comment, thanks!

src/app/models/node.model.ts
Math-R: I added the optional chaining to prevent crashes in existing tests where `addTransitionAndComputeRouting` is called with a `Trainrun` instance missing `trainrunCategory`. It was a quick safeguard to avoid changing the specs. I can revert to the original version and update the test spec instead if you prefer.

louisgreiner: That's a good idea, however, I think that being permissive on the potential missing `trainrun.getTrainrunCategory()` would ease the developping errors above all: `trainrun.getTrainrunCategory()` should never be null.

The error that your code fixes is in this tests `src/app/services/util/port-ordering.algo.spec.ts`:
```sh
Chrome Headless 147.0.0.0 (Windows 10) getConnectedComponents should return single component for connected graph FAILED
        TypeError: Cannot read properties of undefined (reading 'fachCategory')
```

It looks like these lines are missing, after the constructor's calling in the above tests;
https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/blob/4f3a4a7f28ca9eb7b9dcfb4d41ae318e65aeffae/src/app/services/data/trainrun.service.ts#L75-L79

That's a shame it's not done *in* the constructor itself already, and it looks difficult to do so, since the categories and stuff rely on the `DataService`

I dont' have the clue yet, I'll try some things, if you don't come up with a solution before.

emersion: Aha, I see! I haven't got time to look deeply into the issue, but I agree with Louis that it's a bit painful to have tests breaking this assumption. I would prefer to make it so tests don't set up a broken DTO. I can investigate tomorrow if you both don't beat me to it.

louisgreiner: I think the easiest way to do so is to add the `trainrun.setXXX` lines in the incriminated test file directly, since there's no way we can use the `DataService` in the model layer.

Math-R: Does the last changes i made suits to both of you ?
I ended creating a TrainrunCategory designed for tests, and apply it where tr where created without category

louisgreiner: I'm ok with this trade-off

louisgreiner (approved):
Early approve, but this needs to run `npm run format`

Math-R:
> Early approve, but this needs to run `npm run format`

my bad i ran lint and not format :/

done

src/app/models/node.model.ts
```diff
@@ -574,16 +574,18 @@ export class Node {
...
     transition.setPort1Id(port1.getId());
     transition.setPort2Id(port2.getId());
-    transition.setIsNonStopTransit(
-      isNonStop
-        ? true
-        : this.trainrunCategoryHaltezeiten[trainrun.getTrainrunCategory().fachCategory].no_halt,
-    );
+    const defaultIsNonStop =
+      isNonStop === undefined
+        ? (this.trainrunCategoryHaltezeiten[trainrun.getTrainrunCategory().fachCategory]?.no_halt ??
```
emersion: I don't think `this.trainrunCategoryHaltezeiten[trainrun.getTrainrunCategory().fachCategory]` can be null? Do we really need the `?.`?

src/app/models/node.model.ts
```diff
@@ -574,16 +574,18 @@ export class Node {
...
+          false)
+        : isNonStop;
```
emersion: I think this can be simplified, the ternary isn't necessary here:

```ts
transition.setIsNonStopTransit(isNonStop ?? this.trainrunCategoryHaltezeiten[trainrun.getTrainrunCategory().fachCategory].no_halt);
```

emersion (approved):
Two minor comments, feel free to ignore/resolve them.

src/app/models/node.model.ts
Math-R: yes you're right, it was unecessary noise

src/app/models/node.model.ts
Math-R: i made the change using your second suggestion about the ternary

PR #935, opened 2026-04-16
fix(view): hide one-way arrow around hidden nodes

### LINKED ISSUE #934: Filter: Non-stops nodes and one-way arrows doesn't affect filter
### What is the input mode?

Mouse

### What type of issue is it?

Functionality

### What happened?

When one‑way train runs are present, the filter “Display nodes with only stopping trainruns” is applied incompletely in the Netzgrafik.
The affected nodes are filtered correctly and are hidden as expected. However, the corresponding direction arrows representing one‑way train runs are not filtered and remain visible.
This leads to a partially filtered visualization where direction arrows are shown without their related nodes, which can confuse users and result in a misleading representation of the data.

[  ] Display nodes with only stopping trainruns

Image

[x] Display nodes with only stopping trainruns

Image

Filtering should be applied consistently to all related visual elements.
When nodes are hidden due to an active filter, any associated one‑way direction arrows should also be hidden accordingly.

1) load data [networkGraphic (3).json](https://github.com/user-attachments/files/26777999/networkGraphic.3.json)
2) Filter -> Display -> [ ] Display nodes with only stopping trainruns
3) Arrows are visible in Netzgrafik (wrong)

Local

Firefox, Google Chrome, Microsoft Edge, Safari, Opera

2.10.19

louisgreiner:
Close https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/issues/934

### DIFF
--- a/src/app/view/editor-main-view/data-views/trainrunsections.view.ts
+++ b/src/app/view/editor-main-view/data-views/trainrunsections.view.ts
@@ -1133,9 +1133,15 @@ export class TrainrunSectionsView {
         .classed(
           StaticDomTags.TAG_HIDDEN,
           (d: TrainrunSectionViewObject) =>
-            !this.editorView.isTemporaryDisableFilteringOfItemsInViewEnabled() &&
-            (!this.editorView.isFilterDirectionArrowsEnabled() ||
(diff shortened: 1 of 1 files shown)


aiAdrian (approved):
Thank you - very fast fix.

PR #944, opened 2026-04-20
chore: drop two-step CI setup

emersion:
Right now, CI runs in two steps: first the lint stage is run, and only when that completes builds are fired up. This increases CI time quite a bit.

Additionally, format and lint steps are spread across multiple jobs. These jobs are quite fast, however each time we need to checkout sources and install dependencies.

Consolidate all of these jobs in a single one, and stop waiting before starting build jobs.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -16,18 +16,6 @@ permissions:
   deployments: write
 
 jobs:
-  format:
-    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


emersion:
Before this PR, CI took 2m. After this PR, it takes 1m20s.

aiAdrian (approved):
I very like this work

kmer2016 (approved):
Nice!

PR #943, opened 2026-04-20
chore: add GitHub Actions and Docker to Dependabot config

emersion:
This allows Dependabot to open pull requests to update these.

PR #947, opened 2026-04-21
chore: docker upgrade Node.js from 22.21 to 24.15

simonbucher:
## Changes
- Upgrade Node.js Docker image from `node:22.21-alpine` to `node:24.15-alpine`
- Update `.nvmrc` from `22.21` to `24.15`

Patch number intentionally omitted (matching convention from cdde40f) so that the latest patch is always picked up automatically. Dependabot via #943 handles automated Docker image updates.

**Docker Hub:** https://hub.docker.com/layers/library/node/24.15-alpine/images/sha256-8e2c930fda481a6ec141fe5a88e8c249c69f8102fe98af505f38c081649ea749


aiAdrian (approved):
Thank you - LGTM!

emersion (changes requested):
Node v25 is not a LTS release. I would recommend to stick to LTS releases. Node v24 is a LTS release: https://nodejs.org/en/about/previous-releases

We've seen issues in the past due to a mismatch between the Node version used by Dependabot and the one used locally or in CI. It seems like this got fixed in https://github.com/dependabot/dependabot-core/pull/13576, so it should be safe to upgrade to v24.

See also https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/commit/cdde40fe33947fb41e89f865c42ed6e65593dbb3

simonbucher:
> Node v25 is not a LTS release. I would recommend to stick to LTS releases. Node v24 is a LTS release: https://nodejs.org/en/about/previous-releases
> 
> We've seen issues in the past due to a mismatch between the Node version used by Dependabot and the one used locally or in CI. It seems like this got fixed in [dependabot/dependabot-core#13576](https://github.com/dependabot/dependabot-core/pull/13576), so it should be safe to upgrade to v24.
> 
> See also [cdde40f](https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/commit/cdde40fe33947fb41e89f865c42ed6e65593dbb3)

 Thanks for the feedback, I updated the pr to v24 lts

.nvmrc
```diff
@@ -1 +1 @@
-22.21
+24.15.0
```
emersion, line 1: The patch number (last part) was dropped on purpose in https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/commit/cdde40fe33947fb41e89f865c42ed6e65593dbb3 because it ended up being outdated pretty quickly and patch releases will not break our build.

I would be fine keeping the patch number if we have an automated mechanism to update it. https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/pull/943 adds the Docker part, but `.nvmrc` is left unaddressed.

Did you re-add the patch number on purpose? If so, what is your motivation? (Reproducibility perhaps?)

emersion:
@simonbucher Do you think you'll have time to update this PR? It would be nice to merge it :)

simonbucher:
Thanks @emersion for the feedback! I've updated the PR

emersion (approved):
Thanks a lot!

PR #948, opened 2026-04-21
fix: drop extraneous "browser" directory in dist/

emersion:
### fix: drop extraneous "browser" directory in dist/

Angular v18 now places build artifacts in a "browser" subdirectory in the dist/ directory, so that SSR files are not mixed up with browser files. We have no use for SSR, so let's drop that extraneous directory.

This fixes deployment on GitHub Pages and PR previews.

### chore: harmonize dist directory layout

Sometimes we were using dist/ directly, sometimes we were using dist/netzgrafik-frontend/.

Use dist/ everywhere since we have a single project.

### DIFF
--- a/angular.json
+++ b/angular.json
@@ -18,7 +18,7 @@
           "builder": "@angular-devkit/build-angular:application",
           "options": {
             "outputPath": {
-              "base": "dist/netzgrafik-frontend",
+              "base": "dist",
               "browser": ""
             },
             "index": "src/index.html",
@@ -40,9 +40,6 @@
           },
           "configurations": {
             "production": {
-              "outputPath": {
-                "base": "dist"
-              },
               "fileReplacements": [
                 {
                   "replace": "src/environments/environment.ts",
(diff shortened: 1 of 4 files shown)


achrafmohye (commented):
LGTM thanks :)

achrafmohye (approved):

aiAdrian (approved):
LGTM! Let's test

angular.json
```diff
@@ -39,9 +40,6 @@
           },
           "configurations": {
             "production": {
-              "outputPath": {
-                "base": "dist"
-              },
```
louisgreiner, line 44: I tried to look for this information (https://angular.dev/reference/configs/workspace-config), but I could not find why these lines have been dropped?

louisgreiner (approved):
LGTM, tested a few build things and nothing broke.

angular.json
emersion: I've dropped them because they are now identical to the base `outputPath` defined in the parent options.

`configurations` inherits all parent options and overwrites some of them, but there's no point in overwriting with the same `outputPath.base`.

louisgreiner: Thanks a lot for the explanations 👌 LGTM

kmer2016 (approved):
LGTM

PR #949, opened 2026-04-22
fix: Reported issue fixed due of wrong halteZeit (stopping time logic)

### LINKED ISSUE #946: Pass‑through setting hides dwell time in network graph but still applies it internally
### What is the input mode?

Mouse

### What type of issue is it?

Functionality

### What happened?

When trains are routed through a node where dwell times are defined for the respective train category but overridden by the setting “pass-through” (see right), the following behaviour occurs:

In the network graph, the new train is displayed as if it were a pure pass-through.

However, the times in the neighbouring nodes become inconsistent, because internally a dwell time is still applied.

The dwell time in the node is nevertheless stored — it appears in the pearl chain, track occupancy, and takt clock — but it is not visible in the network graph.

This results in an inconsistency between the graphical representation and the time data actually used by the system.

_No response_

(1) BUG -> see stopping times : NO HALT -> bug - should be non-stop -> 0 minute 

https://github.com/user-attachments/assets/04337908-4af1-45ef-ba2a-7c8c6c682e26

(2) BUG -> see stopping times : HALT (stop) -> correct

https://github.com/user-attachments/assets/a2c7ae24-ac85-4292-ace2-117f6cc3e91d

Local

Firefox, Opera, Safari, Microsoft Edge, Google Chrome

2.10.18

aiAdrian:
The reported issue is fixed

https://github.com/user-attachments/assets/5a8e2d09-e5af-4977-b698-ec675c62ed0d

### DIFF
--- a/src/app/services/data/trainrunsection.service.ts
+++ b/src/app/services/data/trainrunsection.service.ts
@@ -76,13 +76,14 @@ export class TrainrunSectionService implements OnDestroy {
   static computeArrivalAndDeparture(
     nodeArrival: number,
     trainrunSection: TrainrunSection,
-    nonStop: boolean,
     halteZeiten: TrainrunCategoryHaltezeit,
     precision: number = 10,
   ): DepartureAndArrivalTimes {
-    let haltezeit =
-      halteZeiten[trainrunSection.getTrainrun().getTrainrunCategory().fachCategory].haltezeit;
-    haltezeit = nonStop ? 0 : haltezeit;
+    let haltezeitObject =
+      halteZeiten[trainrunSection.getTrainrun().getTrainrunCategory().fachCategory];
+
+    console.log("computeArrivalAndDeparture", haltezeitObject);
+    const haltezeit = haltezeitObject.no_halt ? 0 : haltezeitObject.haltezeit;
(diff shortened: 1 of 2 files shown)


src/app/services/data/trainrunsection.service.ts
```diff
@@ -76,13 +76,14 @@ export class TrainrunSectionService implements OnDestroy {
...
     halteZeiten: TrainrunCategoryHaltezeit,
     precision: number = 10,
   ): DepartureAndArrivalTimes {
-    let haltezeit =
-      halteZeiten[trainrunSection.getTrainrun().getTrainrunCategory().fachCategory].haltezeit;
-    haltezeit = nonStop ? 0 : haltezeit;
+    const haltezeitObject =
+      halteZeiten[trainrunSection.getTrainrun().getTrainrunCategory().fachCategory];
+
+    console.log("computeArrivalAndDeparture", haltezeitObject);
```
emersion, line 85: This seems like a remnant from a debugging session?

emersion (commented):
Looks good apart from that comment!

src/app/services/data/trainrunsection.service.ts
aiAdrian: Thanks - i have not seen it :-)

emersion (approved):
LGTM!

PR #952, opened 2026-04-22
chore(deps): remove devtools-protocol from dependencies

louisgreiner:
This dependency is removed after running `npm install`.


emersion:
Ah! I think Dependabot did the same thing in https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/commit/bff5a249bf075413df233429a6e5d2002512190c.

louisgreiner:
Oh, indeed, thanks.

PR #962, opened 2026-04-23
fix: fix traffic side in editor view

Uriel-Sautron:
This fix addresses an omission in `data.service.ts` that prevented the traffic side from updating.


louisgreiner (approved):
LGTM, tested

Math-R (approved):

PR #963, opened 2026-04-24
chore(deps): drop explicit devtools-protocol dependency

emersion:
This was added back in 2892b14320b5 ("chore: add explicit devtools-protocol optional dependency") to fix CI for Dependabot PRs. However, having this as a direct dependency induces a lot of noise because devtools-protocol is updated *very* often.
    
Let's see if that Dependabot bug is fixed now.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -23,7 +23,6 @@
         "angular-oauth2-oidc": "^15.0.1",
         "angular-server-side-configuration": "^18.2.0",
         "d3": "^5.16.0",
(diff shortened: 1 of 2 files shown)


emersion:
I don't have good ideas to test this.

emersion:
I suppose we can just try and see? CI now uses Node v24 which should handle optional peer deps better. We can always revert if it breaks things.

louisgreiner (approved):
Let's try and revert if this breaks anything

SharglutDev (approved):

PR #974, opened 2026-05-04
fix: Ensure the warning is non longer option - older gets migrated

### LINKED ISSUE #973: Older Netzgrafik cause an warning issue, when missing backward travel times due of wrong warning check
### What is the input mode?

Mouse

### What type of issue is it?

Functionality

### What happened?

When we open elder Netzgrafik - we can get warning issue because we only check whether the object is non null - but it can as well be undefined. 

``` typescript
/**
 * Represents the time data in TrainrunSectionDto.
 */
export interface TimeLockDto {
  time: number; // minutes [0..60]
  consecutiveTime: number; // automatically updated after any data changes in the application
  lock: boolean; // used to stop the time propagation (forward/backward)
  warning?: WarningDto; // warning - if business logic detects an issue -> set human-readable warning
  timeFormatter?: TimeFormatter; // undefined or object - optional
}
```
The warning element is optional it can be undefined! 

```typescript 
  hasTravelTimeWarning(): boolean {
    return this.travelTime.warning !== null;
  }

  hasBackwardTravelTimeWarning(): boolean {
    return this.backwardTravelTime.warning !== null;
  }

  hasSourceDepartureWarning(): boolean {
    return this.sourceDeparture.warning !== null;
  }

  hasSourceArrivalWarning(): boolean {
    return this.sourceArrival.warning !== null;
  }

  hasTargetDepartureWarning(): boolean {
    return this.targetDeparture.warning !== null;
  }

  hasTargetArrivalWarning(): boolean {
    return this.targetArrival.warning !== null;
  }
```

if the warning objects exists and the warning element gets reset i can as well get the value ```null``` thus we have to handle this correctly.

See description

Technical issue

Local

Firefox

2.10.19

aiAdrian:


louisgreiner:
I cannot reproduce the [bug](https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/issues/973) this should fix. Can you provide such a json?

aiAdrian:
[netzgrafik_rhb.json](https://github.com/user-attachments/files/27404147/netzgrafik_rhb.json)

or
[netzgrafik_bls.json](https://github.com/user-attachments/files/27404152/netzgrafik_bls.json)

louisgreiner (commented):
Thanks for the files, I reproduced the bug.

I'm not a fan of this approve, because it adds a manual migration that adds a lot of noise to the code. I wonder how, in this same file, `backwardTravelTime` sometimes includes a warning and sometimes not.

I will try to think of an alternative solution. Edit: I couldn't find other solution :s

emersion:
Can we instead adjust `hasXXXWarning()` functions to check for null and undefined instead of just for null? e.g.

```ts
  hasTravelTimeWarning(): boolean {
    return Boolean(this.travelTime.warning);
  }
```

louisgreiner:
The other concern I have is that old files have `backwardTravelTime` not having the field `warning` and if we don't migrate the data, importing and exporting such a file will result in a file still not including the `warning` attributes. Should we do something about it?

aiAdrian:
The questions is how we ensure that we are downwards compatible.

louisgreiner:
After some discussion, we could try this: https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/pull/974#issuecomment-4430585082

louisgreiner:
@aiAdrian what do you think of this solution (https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/pull/974#issuecomment-4430585082)? It would avoid having a migration method called each time in the constructor

louisgreiner:
I propose an alternative fix in this PR:
- https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/pull/1225

If this is ok, we can close this one and merge the other one.

PR #982, opened 2026-05-05
fix(perlenkette): do not display asymmetry arrow when trainrun is one-way

louisgreiner:
When a trainrun is non-symmetric:
image

we used to display both arrows (asymmetry AND one-way) in the perlenkette:
image

With this fix:
image


SharglutDev (approved):
Lgtm ✅

Pivouane (approved):
LGTM 🐉

aiAdrian (approved):
LGTM - small fix - but effective

PR #986, opened 2026-05-06
refactor: stammdaten renamed to base data and fixed reported cvs import/export issues

### LINKED ISSUE #938: CSV Export/Import: Time settings are unclear and inconsistent when importing nodes
### What is the input mode?

Mouse

### What type of issue is it?

Functionality

### What happened?

When nodes are imported via CSV, it is not clear which time values are actually imported from the CSV file and which ones are automatically added by the system.
In particular, when a node is exported with the setting “no stop”, the re-import does not behave transparently:

- Instead of keeping the exported values,
- the system appears to add generic or default time values, similar to what happens when a node is newly created via the UI.

As a result, imported node time settings do not reliably reflect the original CSV data.

Image

It should be clearly defined and reproducible which time values are taken from the CSV and which are computed or added by the system
When exporting nodes with “no stop”, no random or generic default times should be added on import
In this case, the system should store zero values internally for the relevant times
CSV import/export should be lossless and predictable

1. Open a project with existing nodes
2. Configure a node with “no stop”
3. Go to Base data → Export base data
4. Export the nodes as a CSV file
5. Go to Base data → Import base data
6. Import the previously exported CSV
7 .Inspect the imported node’s time settings

Local

Firefox, Opera, Safari, Microsoft Edge, Google Chrome

2.10.8

aiAdrian:
- Refactored the entire stammdaten module to baseData for a consistent and language‑neutral data model.
- Unified all CSV exports to English‑only column names to enable international and language‑independent base‑data exchange.
- Fixed issues related to halt (tag on/off) and non‑rounded times, ensuring reliable import/export behavior.
- Ensured that users can now export and re‑import data without schema mismatches or language‑dependent fields.
- Updated the documentation to reflect the new baseData structure and English column schema.
- Updated the example files to match the new format.

I am aware that this PR touches a large number of files, but this was necessary to achieve a clean, consistent, and future‑proof implementation. Most of the changed files are tests which include the baseDataService (BaseDataService.ts) and this enforced to change the tests.

Test: 
[stammdaten_new_format.csv](https://github.com/user-attachments/files/27454891/stammdaten_new_format.csv)


src/app/services/data/data.service.ts
```diff
@@ -276,8 +276,8 @@ export class DataService implements OnDestroy {
     return Object.values(Direction);
   }
 
-  getBPStammdaten(betriebspunktName: string): Stammdaten {
-    return this.stammdatenService.getBPStammdaten(betriebspunktName);
+  getStationCodeBaseData(stationCode: string): BaseData {
+    return this.baseDataService.getStationCodeBaseData(stationCode);
```
louisgreiner: Since `stationCode` refers to `betriebspunktName`, I would suggest to keep "betriebspunktName" for now and migrate from this word to another one in a dedicated PR.

"stationCode" is nice. I have also "shortName" in mind, to be aligned with "fullName". What do you think? (@emersion also)

src/app/services/data/basedata.service.ts
```diff
@@ -0,0 +1,190 @@
...
+      const filterableLabels: string[] = BaseDataService.parseStringArray(row.Labels);
+      const kategorien: string[] = BaseDataService.parseStringArray(row.Category);
+      const bahnhof: string = row.StationName;
+      const create: number = BaseDataService.parseTimeAsFloat(row.Create);
+      const posX: string = row.XCoord;
+      const posY: string = row.YCoord;
+      let position: Vec2D;
+      if (posX !== undefined && posY !== undefined && posX !== "" && posY !== "") {
+        position = new Vec2D(+posX, +posY);
+      }
```
louisgreiner, line 302: For the other reviewers, this has been added when migrating the file `src/app/services/data/stammdaten.service.ts` to this one (`src/app/services/data/basedata.service.ts`)

src/app/services/ui/ui.interaction.service.ts
```diff
@@ -329,17 +330,22 @@ export class UiInteractionService implements OnDestroy {
...
 
-  closeNodeStammdaten() {
-    this.showNodeStammdatenSubject.next(false);
+  closeNodeBaseData() {
+    this.showNodeBaseDataSubject.next(false);
+  }
+
+  // Backward-compatible alias used by older callers.
+  closeBaseData() {
+    this.closeNodeBaseData();
```
louisgreiner: This method is not needed. This is only called here:
https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/blob/cea758552164c20acd678451887363a5b3f9df41/src/app/view/editor-tools-view-component/editor-tools-view.component.ts#L771-L775

src/app/view/editor-tools-view-component/editor-tools-view.component.ts
```diff
@@ -795,7 +771,7 @@ export class EditorToolsViewComponent {
   private processNetzgrafikJSON(netzgrafikDto: NetzgrafikDto) {
     // prepare JSON import
     this.uiInteractionService.showNetzgrafik();
-    this.uiInteractionService.closeNodeStammdaten();
+    this.uiInteractionService.closeBaseData();
```
louisgreiner:
```suggestion
    this.uiInteractionService.closeNodeBaseData();
```

And `uiInteractionService.closeBaseData()` can be deleted.

louisgreiner (commented):
LGTM, only a few comments.

src/app/services/data/data.service.ts
aiAdrian: fully agree - that we should not replace betriebspunktName so far.

emersion:
This PR is difficult to review because it does multiple things at once (renaming, adjusting data format, etc) and it's not properly split into logical commits.

aiAdrian:
Based on feedback from our business owner the col names have changed again.


### Target column names

- StationCode  
- StationName  
- Category  
- Region  

- MinimumStopTime_IPV  
- PassingThroughStation_IPV  

- MinimumStopTime_A  
- PassingThroughStation_A  

- MinimumStopTime_B  
- PassingThroughStation_B  

- MinimumStopTime_C  
- PassingThroughStation_C  

- MinimumStopTime_D  
- PassingThroughStation_D  

- ZAZ (Train dispatching time)  
- ConnectionTime  

- Labels  
- XCoord  
- YCoord  
- Create  

---

### Rationale

The column names were updated to use clearer and more explicit terminology aligned with the business domain. In particular, ambiguous fields like `DwellTime` and `StopFlag` were replaced with `MinimumStopTime` and `PassingThroughStation` to better reflect their actual meaning and reduce misinterpretation. This change ensures consistency, improves data quality, and aligns the schema with the definitions provided by the business owner.

src/app/services/data/data.service.ts
aiAdrian: but the inferface names are declared by our business. thus (see comments later) we should not change this at this stage

src/app/view/editor-tools-view-component/editor-tools-view.component.spec.ts
```diff
@@ -0,0 +1,95 @@
+import {of} from "rxjs";
+import {BaseDataService} from "../../services/data/basedata.service";
+import {EditorToolsViewComponent} from "./editor-tools-view.component";
+
+describe("EditorToolsViewComponent", () => {
```
louisgreiner: I'm not sure to fully understand this test file, especially I'm not sure I would be able to maintain it if it ever should be modified. I guess this part has been generated.

From my point of view this looks dangerous to add this kind of code in the code base, which is already dense and complex.

I'm open to discussion on keeping this file, but this would need more context and reasons why this test is important.

louisgreiner (commented):
This PR is not easy to review, since the commits history does things at the beginning then undo some others afterwards. Also, this PR does more than a thing (rename + fix bug + add tests): maybe this could have been split in different PRs.

I think most of the PR has been generated, but there is a part that I even don't understand and feel unconfortable about, we can discuss in the comment below.

src/app/view/editor-tools-view-component/editor-tools-view.component.spec.ts
aiAdrian: I will again dbl check the test - and decide to remove it or not

aiAdrian:
@louisgreiner - i understand all thecreview findings, i think.nwxt time i should make smaller changes and with more care about review understanability and maintainablity ... looks like copilot makes to many complex changes at once 😉

src/app/view/editor-tools-view-component/editor-tools-view.component.spec.ts
aiAdrian: After longer re-thinking -> removed

louisgreiner (approved):
LGTM then, thanks for taking the comments into account! :)

PR #987, opened 2026-05-06
fix which node is undocked when drag 6+ collapsed nodes dot

Math-R:
I made a small fixup because when there was 5 collasped nodes on a tsvo between to uncollapsed node, when draging an uncollapsed node to another node, this was always the first collapsed node who was undock.

This must happend only if there is 6+ node, not 5+.


SharglutDev (approved):
Lgtm and tested ✅ Works great

louisgreiner (approved):

PR #991, opened 2026-05-07
fix: implemented in such a way the ux expert proposed

### LINKED ISSUE #990: Improve checkbox visual elements - more contrast
### What is the input mode?

Mouse

### What type of issue is it?

Functionality

### What happened?

Hello Adrian, I'm sorry if we kept you waiting following your answer.
It seems to me the current design for the checkbox is to blame for the user's sentiment and preference toward the toggle switch, not the fact that it's a checkbox in itself.
In their checked state, the use of very light grey make them seem as if they were disabled, and are not contrasted enough. This result in a less efficient overall visual structure for the list, hence the second and third reasons you listed.
Checkboxes can be animated too, and as a matter of fact we are adding them for OSRD right now.

The problem I see with toggles in this context is that they are very loud, having more visual presence than the graph itself, which would be a visual hierarchy failure. I understand their appeal, but in the context where they are numerous, they aren't the best fit.

I don't know if you have the possibility to update the checkbox design, but here is what would work better in my opinion, in terms of contrast:

image

https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/pull/855

_No response_

.

Local

Firefox

2.10.19

aiAdrian:
| before | after | 
| --- | --- | 
|  chrome-capture-2026-05-07 (1)  |  chrome-capture-2026-05-07 |

SEE: https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/issues/990#issue-4397725346

### DIFF
--- a/src/app/app.component.scss
+++ b/src/app/app.component.scss
@@ -78,3 +78,9 @@ input[type="number"]::-webkit-inner-spin-button {
   transform: translate(10px, 1px);
   width: 95%;
 }
+
+::ng-deep .sbb-selection-container-checked { 
+  svg polyline {
+    stroke-width: 2px;
+  } 
+}
--- a/src/app/app.component.scss
+++ b/src/app/app.component.scss
@@ -79,6 +79,11 @@ input[type="number"]::-webkit-inner-spin-button {
   width: 95%;
 }
 
+::ng-deep .sbb-selection-checked .sbb-selection-container {
+  border-width: 2px;
+  border-radius: 3px;
+}
+
 ::ng-deep .sbb-selection-container-checked {
   svg polyline {
     stroke-width: 2px;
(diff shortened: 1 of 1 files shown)


src/app/app.component.scss
```diff
@@ -78,3 +78,9 @@ input[type="number"]::-webkit-inner-spin-button {
   transform: translate(10px, 1px);
   width: 95%;
 }
+
```
louisgreiner, line 81: What do you think of also making the box borders wider and the edges rounder?

```suggestion

::ng-deep .sbb-selection-checked .sbb-selection-container {
  border-width: 2px;
  border-radius: 3px;
}

```


| Before | After |
| --- | --- |
| Image  | Image  |

louisgreiner (commented):
Nice! It feels way clearer that way. What about making borders wider also?

louisgreiner (approved):

PR #996, opened 2026-05-11
fix: spin-buttons issue resolved

### LINKED ISSUE #995: input[number] spin-buttons visible in Webkit browsers (Chrome, Edge, Safari)
### What is the input mode?

Mouse

### What type of issue is it?

Other

### What happened?

The current CSS implementation for hiding the increment/decrement arrows (spin-buttons) in numeric input fields is only working correctly in Firefox. In Webkit-based browsers (Chrome, Microsoft Edge, and Safari), the arrows are still visible, which breaks the UI consistency.

_No response_

1. Draw new trainrun 
2. Open trainrun section and edit time 

Local

Google Chrome, Microsoft Edge, Safari, Opera

2.10.19

aiAdrian:
Issue resolved - firefox , edge, and chrome are working

### DIFF
--- a/src/app/app.component.scss
+++ b/src/app/app.component.scss
@@ -63,15 +63,15 @@ sbb-loading-indicator {
   }
 }
 
-// firefox fix
-::ng-deep input[type="number"]::-webkit-outer-spin-button,
-input[type="number"]::-webkit-inner-spin-button {
-  -webkit-appearance: none;
-  margin: 0;
+// firefox and chrome based browsers fix
+::ng-deep input[type="number"] {
+  -moz-appearance: textfield; /* Firefox */
 }
 
-::ng-deep input[type="number"] {
-  -moz-appearance: textfield;
+::ng-deep input[type="number"]::-webkit-outer-spin-button,
+::ng-deep input[type="number"]::-webkit-inner-spin-button {
+  -webkit-appearance: none; /* Webkit-Browser */
+  margin: 0;
 }
 
 .sbb-select.language-selector-menu {


src/app/app.component.scss
```diff
@@ -63,15 +63,15 @@ sbb-loading-indicator {
...
 }
 
-// firefox fix
-::ng-deep input[type="number"]::-webkit-outer-spin-button,
-input[type="number"]::-webkit-inner-spin-button {
-  -webkit-appearance: none;
-  margin: 0;
+// firefox and chrome based browsers fix
+::ng-deep input[type="number"] {
+  -moz-appearance: textfield; /* Firefox */
```
emersion, line 68: [`appearance`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/appearance) is a standard CSS property, the `-moz-` prefix shouldn't be required and it should work on all browsers.

aiAdrian: I don't understand - sorry @emersion  you might can help me to fix it for chrome, edge and firefox.

emersion: Hm - looking at this in more detail, it seems like although `appearance` is standard, it's not enough to hide spin buttons on Chromium-based browsers.…

The diff is hard to read because things are moved around, but the only thing this PR does is adding a missing `::ng-deep` for `::-webkit-inner-spin-button`.

My bad! I don't see a better way to fix it.

emersion (approved):
LGTM!

louisgreiner (approved):

PR #1001, opened 2026-05-11
chore(deps): upgrade angular-oauth2-oidc to v19

emersion:
angular-oauth2-oidc's major versions are synchronized with Angular. We were using angular-oauth2-oidc v15 with Angular v19, which was unsupported.

Release notes indicate no breaking change:
https://github.com/manfredsteyer/angular-oauth2-oidc/releases

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -20,7 +20,7 @@
         "@angular/platform-browser-dynamic": "^19.2.21",
         "@angular/router": "^19.2.21",
         "@sbb-esta/angular": "^19.1.6",
-        "angular-oauth2-oidc": "^15.0.1",
+        "angular-oauth2-oidc": "^19.0.0",
         "angular-server-side-configuration": "^19.0.1",
         "d3": "^5.16.0",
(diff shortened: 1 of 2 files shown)


SharglutDev (approved):
Lgtm ✅ 

Would it be possible to configure dependabot so we bump angular and angular-oauth2 together ?

emersion:
> Would it be possible to configure dependabot so we bump angular and angular-oauth2 together ?

These need to keep the same major version in sync, but minor/patch versions are desynchronized, so I'm not sure it's a good thing to keep them in lockstep. Maybe in an `update-types: ["major"]` group? But these kind of Angular upgrades will need manual intervention anyways, so I don't know if it would help much.

SharglutDev:
> But these kind of Angular upgrades will need manual intervention anyways, so I don't know if it would help much.

It was more of a reminder not to forget to bump this depth when bumping in the angular major version if it can break some things.

Akctarus (approved):

louisgreiner (approved):

PR #1003, opened 2026-05-11
fix: fix metadata operation

Uriel-Sautron:
To pass `operation` events to the host component, we use the `@output` decorator and `observable`. 
Here we add `uiInteractionService.operation` to access the `metadata` operation


emersion (approved):

louisgreiner (approved):
Nice catch!

It looks like the description of the commit is too long (> 100 char) and should include a line break somewhere.

PR #1007, opened 2026-05-12
fix(traffic side): fix unwanted overwrite

Uriel-Sautron:
`activeTrafficSideType` conflicted with `trafficSide` in the DTO, and it hasn't been used

### DIFF
--- a/src/app/services/ui/ui.interaction.service.ts
+++ b/src/app/services/ui/ui.interaction.service.ts
@@ -110,7 +110,6 @@ export class UiInteractionService implements OnDestroy {
   private activeTheme: ThemeBase = null;
   private activeStreckengrafikRenderingType: StreckengrafikRenderingType = null;
   private activeTravelTimeCreationEstimatorType: TravelTimeCreationEstimatorType = null;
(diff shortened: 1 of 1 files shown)


src/app/services/ui/ui.interaction.service.ts
```diff
@@ -303,11 +302,10 @@ export class UiInteractionService implements OnDestroy {
   }
 
   getActiveTrafficSideType(): TrafficSide {
```
louisgreiner, line 304: Nit: what about shortening everything to "TrafficSide"? like `getTrafficSide()` etc.

louisgreiner (approved):
Nice!

src/app/services/ui/ui.interaction.service.ts
Uriel-Sautron: This was done to align with the other functions (`getActiveTravelTimeCreationEstimatorType`, `getActiveStreckengrafikRenderingType`, ...)

louisgreiner: Mmh true, ignore this then.

SharglutDev (approved):

PR #1009, opened 2026-05-12
fix(editor-main-view): clear all objects when loading new DTO

### LINKED ISSUE #851: Error after importing a reticular similar to an existing one and clicking on a node.

emersion:
When a new DTO is loaded, the editor main view will re-use old SVG elements for new objects with the same ID. d3 captures the old model objects, causing some havoc in event handlers: these mix model objects originating from the old and new DTO.

To fix this, clear all model objects when starting to load a new DTO.

Closes: https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/issues/851

### DIFF
--- a/src/app/view/editor-main-view/editor-main-view.component.ts
+++ b/src/app/view/editor-main-view/editor-main-view.component.ts
@@ -34,6 +34,7 @@ import {NoteDialogParameter, NoteDialogType} from "../dialogs/note-dialog/note-d
 import {AnalyticsService} from "../../services/analytics/analytics.service";
 import {Note} from "../../models/note.model";
 import {LogService} from "../../logger/log.service";
+import {DataService} from "../../services/data/data.service";
 import {UndoService} from "../../services/data/undo.service";
 import {CopyService} from "../../services/data/copy.service";
 import {StreckengrafikDrawingContext} from "../../streckengrafik/model/util/streckengrafik.drawing.context";
@@ -72,6 +73,7 @@ export class EditorMainViewComponent implements AfterViewInit, OnDestroy {
     private uiInteractionService: UiInteractionService,
     private noteService: NoteService,
     private analyticsService: AnalyticsService,
+    private dataService: DataService,
     private undoService: UndoService,
     private copyService: CopyService,
     private logService: LogService,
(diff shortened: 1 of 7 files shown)


aschantraine (approved):
Seems to work well, thank you !

louisgreiner (approved):

aiAdrian (approved):
LGTM

src/app/view/editor-main-view/editor-main-view.component.ts
```diff
@@ -552,5 +554,20 @@ export class EditorMainViewComponent implements AfterViewInit, OnDestroy {
...
+      .subscribe((info) => {
+        if (!info.load) return;
+
+        // When starting to load a new graph, clear any existing graph to avoid
+        // clashes between old and new model objects
+        this.editorView.nodesView.displayNodes([]);
+        this.editorView.transitionsView.displayTransitions([]);
+        this.editorView.connectionsView.displayConnections([]);
+        this.editorView.trainrunSectionsView.displayTrainrunSection([]);
+        this.editorView.notesView.displayNotes([]);
```
kmer2016, line 570: nit:
would an `editorView.clearAll()` method make sense here?

Right now the component has to know which 5 sub-views exist inside editorView. If someone adds a 6th view type later, nothing reminds them to clear it here too.

Not blocking, just wondering if it's worth the small refactor while we're here.

kmer2016 (approved):

src/app/view/editor-main-view/editor-main-view.component.ts
aiAdrian: Indeed, this is a good point. For me, both variants are fine—it's up to @emersion whether to change it or not.
As it stands, I can understand it well, and I haven't encountered any duplication issues so far, so it's fine for me.
However, encapsulating this logic might make it clearer.
Perhaps the method could be named something like clearAllInternalData / clearAllEditorData - instead of clearAll .

emersion: The rest of the function right above also lists each individual type of view, so I'd prefer to keep things consistent. These two parts of the function can be refactored together in a future patch if that's deemed desirable.

PR #654, opened 2025-11-27
feat: make nodes collapsible

Math-R:
- [ ] This branch should be rebased against main
- [ ] "import: set node position before calling replaceIntermediateStopWithNode()" should be squashed into a previous commit
- [ ] "Introduce helper to create view object for TrainrunSectionsView tests" should be dropped

Close https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/issues/674

PR #1011, opened 2026-05-12
Fix travelTime and backwardTravelTime textElement

louisgreiner:
Fixes the travel times positions. Bug can only be reproduce in a specific source-target position.

Before fix (bug, positions overlap):

https://github.com/user-attachments/assets/0284cb62-9e93-43fb-b941-9e8edc2a0b95

After fix:

https://github.com/user-attachments/assets/ddff3ac1-03d1-4600-8a5e-a4bc03418a6b

Bug can be reproduced using this: [networkGraphic(9).json](https://github.com/user-attachments/files/27647233/networkGraphic.9.json)

### DIFF
--- a/src/app/models/trainrunsection.model.ts
+++ b/src/app/models/trainrunsection.model.ts
@@ -770,7 +770,6 @@ export class TrainrunSection {
     this.path.textPositions = SimpleTrainrunSectionRouter.placeTextOnTrainrunSection(
       this.pathVec2D,
       this.sourceNode.getPort(this.sourcePortId),
-      !this.areTravelTimesEqual(),
     );
   }
 
--- a/src/app/services/util/trainrunsection.routing.ts
+++ b/src/app/services/util/trainrunsection.routing.ts
@@ -305,7 +305,6 @@ export class SimpleTrainrunSectionRouter {
   static placeTextOnTrainrunSection(
     lineWayPoints: Vec2D[],
     sourcePort: Port,
-    isBackwardTravelTimeDisplayed: boolean,
   ): TrainrunSectionTextPositions {
     const s = lineWayPoints[0];
     const s1 = lineWayPoints[1];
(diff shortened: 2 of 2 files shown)


aiAdrian (approved):

src/app/services/data/trainrun-section-times.service.ts
```diff
@@ -764,6 +764,7 @@ export class TrainrunSectionTimesService {
         new DirectedTrainrunSectionProxy(this.selectedTrainrunSection, direction),
         this.timeStructure,
       );
+      this.trainrunSectionService.updateText();
```
emersion: This is already called from inside `setTimeStructureToSingleTrainrunSection()`, do we need to do it twice?

src/app/services/data/trainrunsection.service.ts
```diff
@@ -909,6 +909,7 @@ export class TrainrunSectionService implements OnDestroy {
       this.trainrunSectionTimesUpdated(pair.trainrunSection);
     }
 
+    this.updateText();
```
emersion: Travel times may be updated for many reasons, e.g. time propagation, transition undocking, and so on. This only covers a narrow case.

Additionally, this is a pretty big hammer - it affects all trainrun sections.

I think we should try to find a better way to fix this, e.g. by removing `isBackwardTravelTimeDisplayed` in `placeTextOnTrainrunSection()`.

emersion (changes requested):

src/app/services/data/trainrun-section-times.service.ts
louisgreiner: Mmh I don't see where this is called in `setTimeStructureToSingleTrainrunSection()`, but it seems that this call can be removed anyway, since I see surprisingly no difference with or without.

emersion: Oh, I mixed up with `setTimeStructureToTrainrunSections()`.

I find it surprising that `setTimeStructureToTrainrunSections()` calls the function, while `setTimeStructureToSingleTrainrunSection()` doesn't.

src/app/services/data/trainrunsection.service.ts
louisgreiner: I totally agree. However, I have no clue on how to proceed elsewise, since the text positions are directly bound with the travel times values. Also, this update logic will be removed in https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/pull/654.

Uriel-Sautron (approved):
Tested, the bug has been fixed

src/app/services/data/trainrunsection.service.ts
louisgreiner: Sorry for the delay, I've come up to this fixup commit (https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/pull/1011/changes/7c6832131f98b71267cb478771fa117dc5a3be9f), following your advice. It works well :)


https://github.com/user-attachments/assets/99c95dbd-cf03-4960-866e-513189ecae60

louisgreiner: Would the fixup commit resolve this thread?

emersion: Definitely!

emersion (approved):
Sweet, LGTM!

PR #1019, opened 2026-05-15
fix(view): apply border-radius change only to checkbox and not radio buttons

louisgreiner:
[This PR](https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/pull/991) introduces border-radius property to `sbb-checkbox`, which are used for both normal checkboxes and radio buttons.

The changes only apply this property to checkboxes.

|   Before  |  After   |
| --- | --- |
|   image  |   image  |

### DIFF
--- a/src/app/app.component.scss
+++ b/src/app/app.component.scss
@@ -79,13 +79,13 @@ sbb-loading-indicator {
   width: 95%;
 }
 
-::ng-deep .sbb-selection-checked .sbb-selection-container {
+::ng-deep .sbb-checkbox .sbb-selection-container {
   border-width: 2px;
   border-radius: 3px;
-}
 
-::ng-deep .sbb-selection-container-checked {
-  svg polyline {
-    stroke-width: 2px;
+  &-checked {
+    svg polyline {
+      stroke-width: 2px;
+    }
   }
 }


src/app/app.component.scss
```diff
@@ -81,6 +81,9 @@ sbb-loading-indicator {
 
 ::ng-deep .sbb-selection-checked .sbb-selection-container {
   border-width: 2px;
+}
+
+::ng-deep .sbb-checkbox-group-vertical .sbb-selection-checked .sbb-selection-container {
```
emersion: Using `.sbb-checkbox-group-vertical` will only target some checkboxes, not all. Shouldn't we use `.sbb-checkbox` instead?

src/app/app.component.scss
```diff
@@ -81,6 +81,9 @@ sbb-loading-indicator {
 
 ::ng-deep .sbb-selection-checked .sbb-selection-container {
   border-width: 2px;
```
emersion: https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/pull/991 has added this rule - should we also make sure it only affects checkboxes and not radio buttons?

src/app/app.component.scss
emersion: What about this rule? Do we also need to narrow down the selector?

src/app/app.component.scss
louisgreiner: I prefer the radio buttons with this `border-width` property than the ones before. If you are ok we can keep it like this?

|  Without property   | With property    |
| --- | --- |
|   image  |  image   |

src/app/app.component.scss
louisgreiner: For checkboxes:

|  With `stroke-width: 2px`  | Without    |
| --- | --- |
|   image  |  image   |

For radio-buttons:

|  With `stroke-width: 2px`  | Without    |
| --- | --- |
|  image   |  image  |

I see no impact for the radio-button (since the radio-buttons do not generate an svg as for the checkboxes), but I guess we could narrow it just in case.

src/app/app.component.scss
louisgreiner: Indeed, this works with `.sbb-checkbox`, fixuped.

src/app/app.component.scss
emersion: I would prefer to narrow it. Maybe a sbb-esta upgrade will end up using SVGs for radio buttons.

src/app/app.component.scss
emersion: I'd prefer to keep deviations from sbb-esta as low as possible, because they make it so NGE is less consistent with other SBB webapps and may induce breakage during dep upgrades (we're touching sbb-esta's internals here).

That said, if you think it's important and Adrian prefers it as well, I'd be fine with it.

louisgreiner: I actually hadn't this in mind, thanks. Would the [last fixup](https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/pull/1019/commits/1a99954a3e9e3a18de35c321d456b52a05efc803) be ok for you? I also removed the repetition of `.sbb-checkbox`.

src/app/app.component.scss
```diff
@@ -79,16 +79,15 @@ sbb-loading-indicator {
...
-
-::ng-deep .sbb-checkbox .sbb-selection-checked .sbb-selection-container {
-  border-radius: 3px;
-}
-
-::ng-deep .sbb-checkbox .sbb-selection-container-checked {
-  svg polyline {
-    stroke-width: 2px;
+::ng-deep .sbb-checkbox {
+  .sbb-selection-container {
```
emersion: Nit: I think these two can be grouped under a single rule?

emersion (approved):
LGTM, feel free to resolve/ignore the comment.

DucNg (approved):
Tested it's working, also nice refactoring

PR #1020, opened 2026-05-15
fix: compute OD matrix for a full day (#922)

### LINKED ISSUE #922: OD Matrix : Some cells say there is no path between two nodes, but there is one
### What is the input mode?

Mouse

### What type of issue is it?

Functionality

### What happened?

I imported this network graph : 

[reticulaire (49).json](https://github.com/user-attachments/files/26602679/reticulaire.49.json)

And opened the OD Matrix/ One can notice for example that there is no path from Lyon to Berlin but there is a path from Berlin to Lyon. 

There should be a path in both directions since there are only roundtrips in this network graph.

Just import the previous file and open the OD matrix

Local, Embedded (like in OSRD)

Firefox, Google Chrome

2.10.19

shenriotpro:
fixes https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/issues/922

Performance-wise, I can't tell the difference as a user.


louisgreiner (approved):
Let's try this and see if users notice a difference

emersion (approved):

PR #1025, opened 2026-05-18
fix: Ensure that only correct/valid (feasible) connections gets added

### LINKED ISSUE #839: Connections in the “pearl chain” are ignoring the one‑way trainrun direction
### What is the input mode?

Mouse

### What type of issue is it?

Functionality

### What happened?

Currently, in the pearl‑chain view, connections are shown as if they worked in both directions.
For users, it is not visible that some connections only exist in one direction.

Main train in the pearl chain → correctly visible

Connecting trains → must also be shown in the pearl chain as directed one 

Add the arrow symbol for each trainrun in the connection view - add the one-way arrow to  all trainruns un the connection view to improve readability (default: roundtrip)

Image

When the trainrun operates only in one-way (one-direction) -> hide the trainrun which is not operating 

Image

https://github.com/user-attachments/assets/c6de962d-c40a-448c-a591-388a4c0c64f3

Local

Firefox

2.10.15

aiAdrian:
Changed
------------

- Ensure only feasible connection can be added  -> node.service 
- Ensure the UI behaves correctly - user can only draw feasible connections -> node.view

Example: [netzgrafik (5).json](https://github.com/user-attachments/files/27958579/netzgrafik.5.json)
 
chrome-capture-2026-05-18

https://github.com/user-attachments/assets/a862a251-b028-4c86-8211-55a4c9653a9f


aiAdrian:
I've not yet included the data‑migration step for loading a Netzgrafik created with an older version, where users were able to draw all connections freely.

We also have not defined how to handle the switch from round‑trip mode (the common case) to one‑way mode. In this situation, all infeasible connections must be removed.

> This requires discussion — we must decide where this correction should be implemented in the architecture.

> https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/issues/839

Either we include this fix as well into this PR / change or we open a new issue and fix it just after we have merge this small fix. => What should we do?

src/app/view/editor-main-view/data-views/nodes.view.ts
```diff
@@ -903,7 +902,21 @@ export class NodesView {
             if (obj.attr(StaticDomTags.EDGE_NODE_ID) === "" + node.getId()) {
               if (obj.attr(StaticDomTags.EDGE_ID) !== "" + ts.getId()) {
                 if (obj.attr(StaticDomTags.EDGE_LINE_LINE_ID) !== "" + ts.getTrainrunId()) {
+                  // general case: only allow connection if the trainrun section belongs
+                  //  to a different trainrun than the dragged trainrun section
                   obj.classed(StaticDomTags.EDGE_LINE_PIN_CONNECTION, hover);
```
emersion: This call sets the class, but it's overwritten right after with a different value. I think we should remove this line?

src/app/view/editor-main-view/data-views/nodes.view.ts
```diff
@@ -22,7 +22,6 @@ import {NodeViewObject} from "./nodeViewObject";
 import {ConnectionsView} from "./connections.view";
 import {EditorMode} from "../../editor-menu/editor-mode";
 import {LevelOfDetail} from "../../../services/ui/level.of.detail.service";
-
```
emersion, line 25: Nit: missing blank newline between imports and class.

emersion (commented):
Overall looks good, just two minor comments!

src/app/view/editor-main-view/data-views/nodes.view.ts
aiAdrian: Good catch! Thanks for pointing that out. I have added the blank newline between the imports and the class definition to clean up the code.

src/app/view/editor-main-view/data-views/nodes.view.ts
aiAdrian: Thank you .... fixed

aiAdrian:
> https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/pull/1025#issuecomment-4478139627

Thank you for the review. I suggest when ok i will merge the change but we still have the above open question.

emersion (approved):
LGTM!

louisgreiner (approved):
In response of https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/pull/1025#issuecomment-4478139627

LGTM, merging this PR and fix in another one the missing feasability check for data migration sounds reasonable to me.

emersion:
Yeah, can we open separate issues to discuss these two points you raised? These are good points!

Issue #706, opened 2025-12-19
[Feature request]: Add a Content Security Policy


emersion:
Create a [Content Security Policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/CSP) for NGE and embed it in a `<meta>` tag.

This would mitigate issues such as https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/pull/702.

aiAdrian:
The proposed change sounds reasonable. However, we need to ensure that the change does not affect the connection to the backend (Netzgrafik Editor) or cause any connection or communication issues. At the moment, I do not expect any risks here, but we should keep this in mind.

Afterwards, everything should be thoroughly tested against OSRD, as a JavaScript API is used there. This is where I see a potential risk of unintended side effects that could be triggered by the adjustment.

Nevertheless, I suggest proceeding with this change with high priority, as it increases overall software and system security.

We also need to consider the dependency that we must not interfere with the ongoing testing and bugfixing for version 2.10.x in the coming days. Therefore, we should evaluate whether it would be better to release this change with version 2.11 instead. However, I lack the experience to assess whether this could introduce potential risks for version 2.10, which is planned to go productive by mid‑February 2026.

PR #1029, opened 2026-05-19
feat: add strict Content-Security-Policy for standalone mode

emersion:
Content-Security-Policy can help mitigate XSS injections. They restrict where content can be loaded from by the browser. More info: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy

This policy would've helped reduce the impact of the following XSS vulnerability:
https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/pull/702

The policy is applied to the standalone mode only for now, to not regress production deployments which may need to phone internal backends hosted at a different origin.

The policy is as strict as possible without making code changes. The browser will print an error to the console when the webapp tries to load something forbidden by the policy.

The policy needs to be relaxed in a few spots:

- SBB icons are loaded by JavaScript from icons.app.sbb.ch. This requires this domain name to be added to connect-src, and data: URLs to be added to img-src.
- Fonts and styles are loaded from cdn.app.sbb.ch.
- Some CSS style is generated by JavaScript and set via the style HTML/SVG attribute. This requires 'unsafe-inline' to be added to style-src.

References: https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/issues/706

### DIFF
--- a/src/environments/standalone/index.html
+++ b/src/environments/standalone/index.html
@@ -2,6 +2,18 @@
 <html id="NetzgrafikRootHtml" lang="en" class="sbb-lean">
   <head>
     <meta charset="utf-8" />
+    <meta
+      http-equiv="Content-Security-Policy"
+      content="
+        default-src 'self';
+        frame-src 'none';
+        object-src 'none';
+        connect-src 'self' https://icons.app.sbb.ch;
+        img-src 'self' data: https://icons.app.sbb.ch;
+        font-src 'self' https://cdn.app.sbb.ch;
+        style-src 'self' 'unsafe-inline' https://cdn.app.sbb.ch;
+      "
+    />
     <title>Netzgrafik-Editor</title>
     <meta name="viewport" content="width=device-width, initial-scale=1" />
   </head>


Signez (approved):
Makes sense to me, as it is a good practice across the industry as a _defense in depth_ security approach and it's quite low risk to add (it's easy to override by a fork and/or to revert if downstream users are frustrated with that change).

aiAdrian:
What is the impact on our in‑house hosted application and its backend? Did you test it against a local backend installation to ensure we don’t run into unexpected behaviour, or should I retest it on my local machine?

emersion:
> What is the impact on our in‑house hosted application and its backend? Did you test it against a local backend installation to ensure we don’t run into unexpected behaviour, or should I retest it on my local machine?

From the PR description:

>The policy is applied to the standalone mode only for now, to not regress production deployments which may need to phone internal backends hosted at a different origin.

In other words: this PR doesn't have any impact on the hosted version.

aiAdrian:
I tested it with backend on my local engine - it looks good.

aiAdrian (approved):
LGTM!

emersion:
Thanks for testing!

Math-R (approved):
Makes sense to me. CSP is a good additional security layer, and the added exceptions for icons/CDN assets and inline styles seem reasonable given the current implementation.

PR #1031, opened 2026-05-20
fix(view): don't use attr() to set js executing attributes

Khoyo:
This doesn't work when using a strict CSP, such as the one introduced by ea5611fb9d0e5. To reproduce, start the app in standalone mode and try to right-click select.

### DIFF
--- a/src/app/view/knoten-auslastung-view/knoten-auslastung-view.component.ts
+++ b/src/app/view/knoten-auslastung-view/knoten-auslastung-view.component.ts
@@ -83,9 +83,9 @@ export class KnotenAuslastungViewComponent implements AfterViewInit, OnDestroy {
   }
 
   ngAfterViewInit(): void {
-    this.svgDrawingContext = d3
-      .select("#knotenAuslastungContainer")
-      .attr("oncontextmenu", "return false;");
+    this.svgDrawingContext = d3.select("#knotenAuslastungContainer").on("contextmenu", () => {
+      d3.event.preventDefault();
+    });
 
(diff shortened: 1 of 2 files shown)


src/app/view/knoten-auslastung-view/knoten-auslastung-view.component.ts
```diff
@@ -83,9 +83,9 @@ export class KnotenAuslastungViewComponent implements AfterViewInit, OnDestroy {
   }
 
   ngAfterViewInit(): void {
-    this.svgDrawingContext = d3
-      .select("#knotenAuslastungContainer")
-      .attr("oncontextmenu", "return false;");
+    this.svgDrawingContext = d3.select("#knotenAuslastungContainer").on("contextmenu", () => {
```
emersion, line 86: The event object should be provided as the callback argument. Could we use that instead of a global variable?

Khoyo: Not on d3 v5 ;)

Khoyo: See the v6 migration guide: https://observablehq.com/@d3/d3v6-migration-guide#events

emersion: Ah yeah, we're stuck on a very old d3 version 😥

emersion (approved):
Good catch, LGTM!

PR #1039, opened 2026-05-22
chore(github): bump Dependabot max PR limit

emersion:
Dependabot stops opening PRs when 5 are already open: https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#open-pull-requests-limit-

This is not enough: we have lots of dependencies and some PRs may be blocked because they require manual action. Still, we don't want to loose other updates.

Use the same limit as OSRD.


louisgreiner (approved):
Thanks again for all the chore's work!

PR #1073, opened 2026-05-26
chore(deps): add missing papaparse types

emersion:
papaparse doesn't ship its own TypeScript definitions. We need to install a separate package to grab them.

Remove ParseResult because parse() already declares its return type and it's invalid:

    src/app/view/editor-tools-view-component/editor-tools-view.component.ts:184:26 - error TS2314: Generic type 'ParseResult<T>' requires 1 type argument(s).

    184       const finalResult: ParseResult = parse(reader.result.toString(), {
                                 ~~~~~~~~~~~

    src/integration-testing/stammdaten.service.test.spec.ts:83:24 - error TS2314: Generic type 'ParseResult<T>' requires 1 type argument(s).

    83     const finalResult: ParseResult = parse(stammdatenCSV, {header: true});
                          ~~~~~~~~~~~

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -44,6 +44,7 @@
         "@types/jasmine": "~3.6.11",
         "@types/jasminewd2": "~2.0.13",
         "@types/node": "^24.12.3",
+        "@types/papaparse": "^5.3.2",
         "angular-eslint": "^19.8.1",
         "browserstack-local": "^1.5.12",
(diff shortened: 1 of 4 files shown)


theocrsb (approved):
lgtm (not tested)

louisgreiner (approved):

SarahBellaha (approved):

PR #1074, opened 2026-05-26
refactor: add missing d3 types

### LINKED ISSUE #678: [TECH] Fix d3 type errors
Currently type definitions for d3 aren't installed. All d3 types are marked as `any`, so type-checks are disabled.

Install d3 types and fix all of the type issues:

```sh
npm install --save-dev @types/d3
npm run build # ouch, so many errors
```

emersion:
_See individual commits._

d3 packages don't provide their own type definitions. We need to grab them via @types/d3.

A handful of preliminary commits fix the resulting ~180 type errors.

Closes: https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/issues/678

### DIFF
--- a/src/app/view/editor-main-view/data-views/connections.view.ts
+++ b/src/app/view/editor-main-view/data-views/connections.view.ts
@@ -123,7 +123,7 @@ export class ConnectionsView {
     return false;
   }
 
-  createConnectionCurve(drawingGroup: d3.selector) {
+  createConnectionCurve(drawingGroup: d3.Selection<SVGElement, undefined, HTMLElement, undefined>) {
     drawingGroup
       .append(StaticDomTags.CONNECTION_LINE_SVG)
       .attr("class", StaticDomTags.CONNECTION_LINE_CLASS)
@@ -148,7 +148,10 @@ export class ConnectionsView {
       );
   }
 
-  createConnectionSinglePin(drawingGroup: d3.selector, pinPos: Vec2D) {
+  createConnectionSinglePin(
+    drawingGroup: d3.Selection<SVGElement, undefined, HTMLElement, undefined>,
+    pinPos: Vec2D,
+  ) {
     const draggable = d3
       .drag()
       .on("start", (cv: ConnectionsViewObject, i, a) =>
@@ -190,7 +193,7 @@ export class ConnectionsView {
       );
   }
 
-  createConnectionPins(drawingGroup: d3.selector) {
+  createConnectionPins(drawingGroup: d3.Selection<SVGElement, undefined, HTMLElement, undefined>) {
     const selectedTrainrun = this.editorView.getSelectedTrainrun();
 
     drawingGroup.each((c: ConnectionsViewObject, i, a) => {
--- a/src/app/view/editor-main-view/data-views/static.dom.tags.ts
+++ b/src/app/view/editor-main-view/data-views/static.dom.tags.ts
@@ -46,9 +46,9 @@ export class StaticDomTags {
   /*
        GROUP object
      */
-  static SVG = "svg";
-  static GROUP_SVG = "g";
-  static GROUP_DOM_REF = "g";
+  static SVG = "svg" as const;
+  static GROUP_SVG = "g" as const;
+  static GROUP_DOM_REF = "g" as const;
 
   /*
       NODE (root group)
@@ -59,7 +59,7 @@ export class StaticDomTags {
        |- NODE_LABELAREA_TEXT        : text in-front of the node label are (drag node)
        |- NODE_CONNECTIONTIME_TEXT   : text in-front of the node connection time open detail view dialog on the right
      */
-  static NODE_SVG = "g";
+  static NODE_SVG = "g" as const;
   static NODE_ROOT_CONTAINER = "root_container_nodes";
   static NODE_ROOT_CONTAINER_DOM_REF = "g.root_container_nodes";
   static NODE_CLASS = "node";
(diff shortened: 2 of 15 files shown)


src/app/view/editor-main-view/data-views/connections.view.ts
```diff
@@ -148,7 +148,10 @@ export class ConnectionsView {
       );
   }
 
-  createConnectionSinglePin(drawingGroup: d3.selector, pinPos: Vec2D) {
+  createConnectionSinglePin(
+    drawingGroup: d3.Selection<SVGElement, undefined, HTMLElement, undefined>,
```
Caracol3, line 152: nit: could `SVGElement` be more precise here? Not sure it's worth the effort though

src/app/view/editor-main-view/data-views/static.dom.tags.ts
```diff
@@ -59,7 +59,7 @@ export class StaticDomTags {
        |- NODE_LABELAREA_TEXT        : text in-front of the node label are (drag node)
        |- NODE_CONNECTIONTIME_TEXT   : text in-front of the node connection time open detail view dialog on the right
      */
-  static NODE_SVG = "g";
+  static NODE_SVG = "g" as const;
   static NODE_ROOT_CONTAINER = "root_container_nodes";
   static NODE_ROOT_CONTAINER_DOM_REF = "g.root_container_nodes";
   static NODE_CLASS = "node";
```
louisgreiner, line 65: This is still not clear to me why we don't we add `as const` to all the other static properties of `StaticDomTags`?

src/app/view/editor-main-view/data-views/trainrunsection.previewline.view.ts
```diff
@@ -60,13 +60,15 @@ export class TrainrunSectionPreviewLineView {
     private versionControlService: VersionControlService,
   ) {}
 
-  static setGroup(nodeGroup: d3.Selector) {
+  static setGroup(nodeGroup: d3.Selection<SVGElement, undefined, HTMLElement, undefined>) {
```
louisgreiner, line 63: In `src/app/view/editor-main-view/data-views/nodes.view.ts`, `nodeGroup` is defined without `HTMLElement` (`null` instead): 

```ts
  setGroup(nodeGroup: d3.Selection<SVGElement, undefined, null, undefined>) {
```

What is the added value of `HTMLElement` here?

Some other classes use this, some others not.

src/app/view/util/svg.mouse.controller.ts
```diff
@@ -65,7 +65,7 @@ export class SVGMouseController {
 
   init(viewboxProperties: ViewboxProperties) {
     this.viewboxProperties = viewboxProperties;
-    const element = d3.select("html").node();
+    const element = d3.select<HTMLElement, undefined>("html").node();
```
louisgreiner, line 68: Nit: Should we start preventing this to be `null`? same in `src/app/view/util/generalViewFunctions.ts`

```suggestion
    const element = d3.select<HTMLElement, undefined>("html").node()!;
```

src/app/view/editor-main-view/data-views/trainrunsections.view.ts
```diff
@@ -137,7 +137,7 @@ export class TrainrunSectionsView {
       .innerRadius(0)
       .startAngle(-Math.PI / 2 + rotate)
       .endAngle(Math.PI / 2 + rotate);
-    return arcGenerator();
+    return arcGenerator(undefined);
```
louisgreiner, line 140: My IDE warns that your change is not supported.

```
No overload matches this call.
  Overload 1 of 2, '(this: any, d: DefaultArcObject, ...args: any[]): string | null', gave the following error.
    Argument of type 'undefined' is not assignable to parameter of type 'DefaultArcObject'.
  Overload 2 of 2, '(this: any, d: DefaultArcObject, ...args: any[]): void', gave the following error.
    Argument of type 'undefined' is not assignable to parameter of type 'DefaultArcObject'.ts(2769)
```

This aligns the type given in parameter and also make sure we return a `string`.

```suggestion
    return arcGenerator(undefined as unknown as d3.DefaultArcObject) || "";
```

I'm not sure of all of that tho.

louisgreiner (commented):
Just a few comments, otherwise the application seems to work the same as before.

src/app/view/editor-main-view/data-views/connections.view.ts
emersion: You mean using `SVGElement` instead of `HTMLElement`?

The `d3.select()` function returns either `Selection<GElement, OldDatum, HTMLElement, any>` or `Selection<GElement, OldDatum, null, undefined>`. The third generic parameter is the parent element. I don't think `SVGElement` is valid TypeScript here. The root element is `HTMLElement` so I think it still makes sense to use that.

We could use `null` instead, if you'd prefer. That would be quite a large change and I'm not sure about the upsides, so I'd wait for feedback from you before doing anything.

src/app/view/editor-main-view/data-views/static.dom.tags.ts
emersion: Have you seen the commit message from "refactor(view): narrow type for SVG element name constants"? If so, I can try to explain in more detail.

Other properties don't help with d3's `append()` type safety: CSS selectors can't be statically analyzed at compile-time to figure out the type of the return element.

src/app/view/editor-main-view/data-views/trainrunsection.previewline.view.ts
emersion: See the discussion above, feel free to comment there: https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/pull/1074#discussion_r3309784972

Hm, I suppose that since strict null checks are disabled project-wide, TypeScript is fine with converting `HTMLElement` to `null`. It's a bit difficult to figure out the implications with that option disabled…

src/app/view/editor-main-view/data-views/connections.view.ts
emersion: Oh, actually, `null` doesn't seem to be valid when `d3.select()` is used with a `string` as input. It's only valid when used with an element as input.

```ts
export function select<GElement extends BaseType, OldDatum>(
    selector: string,
): Selection<GElement, OldDatum, HTMLElement, any>;

export function select<GElement extends BaseType, OldDatum>(
    node: GElement,
): Selection<GElement, OldDatum, null, undefined>;
```

So I don't think `null` would be valid for NGE.

src/app/view/util/svg.mouse.controller.ts
emersion: This PR is only about adding types and making them work well with NGE's current TypeScript configuration, which disables strict null checks. It seems your editor doesn't use the project's TypeScript configuration.

I'd prefer to address null issues in separate PRs, since it's a much more large chunk of work.

src/app/view/editor-main-view/data-views/trainrunsections.view.ts
emersion: I think this is due to your IDE not using the project's TypeScript configuration. CI doesn't surface such errors.

This double-cast chain is unsafe, I don't think we should add it.

src/app/view/editor-main-view/data-views/static.dom.tags.ts
louisgreiner: Oh it's because of `append()` that is called on specific elements only. Thanks

src/app/view/editor-main-view/data-views/connections.view.ts
emersion: I've pushed a fixup commit to use `HTMLElement` everywhere.

src/app/view/util/svg.mouse.controller.ts
louisgreiner: That is fine by me!

src/app/view/editor-main-view/data-views/connections.view.ts
Caracol3: Actually I was referring to the first parameter. Could `SVGElement` be replaced by something more specific like `SVGGElement` (not sure it's right and worth the change). But your fix is a good catch too, thanks!

emersion: It could be made more specific but nothing in the function requires it to be more specific, so not sure it's worth auditing each function to figure out which exact element we're manipulating.

Caracol3 (approved):
LGTM

louisgreiner (approved):
Thanks a lot!

emersion:
(Trivial rebase onto main.)

kmer2016 (approved):

emersion:
Thanks for the detailed reviews!

PR #1078, opened 2026-05-27
fix(model): initialize potential missing TrainrunSection.path

louisgreiner:
Also fixes the third-party import by removing the check on TrainrunSection.path which is useless, since this attribute is always recomputed.

This needs to be tested internally at SBB with the third party software.

PR #1080, opened 2026-05-28
fix(service): correctly instantiate TrainrunSection when coming from 3rd party

louisgreiner:
Follow-up of https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/pull/1078

You can test by trying to import this: https://github.com/user-attachments/files/28323164/FV21_Paris-Frankfurt_Viriato_NGE.json

Fixes the third-party import.

Open discussion: it is difficult to maintain the third-party import without being able to use it; how can we do better? Also, I think that the third party import should be updated to be aligned with Netzgrafik-Editor evolutions, as OSRD does.


aiAdrian (approved):
Thank you for this very quick FIX !!!

emersion (approved):

DucNg (approved):

PR #1091, opened 2026-06-01
chore: drop end-to-end test framework

emersion:
Our end-to-end tests were empty and never run. They rely on a framework which no longer works:
https://blog.angular.dev/protractor-deprecation-update-august-2023-2beac7402ce0

Running the tests results in a fatal error message.

We can re-introduce an E2E test framework later if we decide to write real tests.


aiAdrian (approved):
LGTM

Synar (approved):

Math-R (approved):

PR #1116, opened 2026-06-03
chore(deps): drop puppeteer

emersion:
The only reason we have this package as a dependency is a CHROME_BIN env var in the karma config file. I don't think this is useful.

### DIFF
--- a/karma.conf.cjs
+++ b/karma.conf.cjs
@@ -1,10 +1,6 @@
 // Karma configuration file, see link for more information
 // https://karma-runner.github.io/1.0/config/configuration-file.html
 
-if (process.env.BUILD_NUMBER) {
-  process.env.CHROME_BIN = require("puppeteer").executablePath();
-}
-
 module.exports = function (config) {
   config.set({
     basePath: "",
--- a/package-lock.json
+++ b/package-lock.json
@@ -64,9 +64,6 @@
         "ts-node": "~10.9.2",
         "typescript": "^5.8.3",
         "typescript-eslint": "^8.60.1"
-      },
-      "optionalDependencies": {
-        "puppeteer": "^22.13.1"
       }
     },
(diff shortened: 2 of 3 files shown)


louisgreiner (approved):
LGTM, I also tested to launch the test `npm run test` since this uses Chrome (this might be useless since the tests use Chrome Headless, I don't know, anyway, it works)

SarahBellaha (commented):
I tested this change by running `npm run test -- --browsers Chrome` as documented in the README, and it fails on my environment with:

```
ERROR [launcher]: No binary for Chrome browser on your platform.
Please, set "CHROME_BIN" env variable.
```

emersion:
Do you have Chrome installed? Which OS are you using? (On my system, I have Chromium installed and I need to set `CHROME_BIN=/usr/bin/chromium` explicitly.)

I don't think this error is caused by this branch - could you try on the `main` branch?

SarahBellaha:
> Do you have Chrome installed? Which OS are you using? (On my system, I have Chromium installed and I need to set `CHROME_BIN=/usr/bin/chromium` explicitly.)
> 
> I don't think this error is caused by this branch - could you try on the `main` branch?

I'm on Ubuntu 24.04 and I have Chromium installed. However, I hadn't explicitly set the `CHROME_BIN` variable before, and it works great now that I have :)
Would it be worth it to add a Chrome/Chromium binary detection to the karma configuration ?

SarahBellaha (approved):

emersion:
> Would it be worth it to add a Chrome/Chromium binary detection to the karma configuration ?

This [used to be the case](https://github.com/karma-runner/karma-chrome-launcher/commit/46e14284ef9bd575ec88b9e62f36f81aaedd56d2), but karma has then split Chromium into its own browser entry. `npm run test -- --browsers Chromium` should work though.

PR #1122, opened 2026-06-05
chore(dependabot): fix update-types type

emersion:
Dependabot wants an array here:

> The property '#/updates/0/ignore/0/update-types' of type string
> did not match the following type: array

Regressed in https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/pull/1115


louisgreiner (approved):

Math-R (approved):
lgtm

PR #1123, opened 2026-06-05
chore(deps): upgrade d3 to v6

emersion:
_See individual commits. Unfortunately I haven't found a good way to split the last commit into smaller chunks._

Migration guide: https://observablehq.com/@d3/d3v6-migration-guide

The following changes were required to migrate:

- Add an event argument to all .on() listeners, use it instead of the d3.event
  global
- Use event.currentTarget instead of a[i] in listeners (a new
  getMouseEventCurrentTarget() helper has been introduced to retain type
  safety)
- Replace d3.mouse() with d3.pointer() (this requires passing around the
  MouseEvent)

d3.drag() event listeners use the custom D3DragEvent event. The drag start
listener can access the current DOM element via event.sourceEvent.currentTarget.
However drag move and end listeners can't do that: currentTarget is set to the
Window object in that case. We need to store the DOM element when drag starts
to be able to access them during move/end.

Some event handlers passed a null domObj to functions down the stack (e.g.
hoverNode() and unhoverNode()). When domObj is non-null, these functions raise
and apply CSS class changes to the element. Add an explicit parameter to
control this behavior now that domObj is carried by the event parameter.

~~Depends on https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/pull/1158~~

TODO:

- [x] Double-check `null` vs `a[i]` in event handlers
- [x] Fix drag event `currentTarget` being `Window`
- [x] Add helper to extract `SVGElement` from `MouseEvent`


src/app/view/editor-main-view/data-views/editor.view.ts
```diff
@@ -708,15 +708,19 @@ export class EditorView implements SVGMouseControllerObserver {
...
-  onGraphContainerMouseup(mousePosition: Vec2D, onPanning: boolean) {
+  onGraphContainerMouseup(event: MouseEvent, mousePosition: Vec2D, onPanning: boolean) {
     if (this.isMultiSelectOn) {
       return;
     }
     if (!this.trainrunSectionPreviewLineView.isDragging() && !onPanning) {
+      const domObj = event.target;
+      if (!(domObj instanceof SVGElement)) {
+        throw new Error("Mouse event's target is not an SVG element");
+      }
```
louisgreiner, line 719: This looks very similar to `getMouseEventCurrentTarget()` from `d3.utils.ts`.

This check on `MouseEvent.target` is done thrice in this commit. Could this be extracted as a static method in `d3.utils.ts`?

src/app/view/util/svg.mouse.controller.ts
```diff
@@ -287,62 +284,66 @@ export class SVGMouseController {
...
+    event.preventDefault();
+    event.stopPropagation();
   }
 
-  private onMouseDownZoom() {
-    const d = d3.select(d3.event.target);
+  private onMouseDownZoom(event: MouseEvent) {
+    if (!(event.target instanceof SVGElement)) {
+      throw new Error("Mouse event's target is not an SVG element");
+    }
```
louisgreiner, line 335: cf. comment about `MouseEvent.target` check.

src/app/view/util/svg.mouse.controller.ts
```diff
@@ -353,26 +354,29 @@ export class SVGMouseController {
...
+    event.preventDefault();
+    event.stopPropagation();
   }
 
-  private onDblclick() {
-    const d = d3.select(d3.event.target);
+  private onDblclick(event: MouseEvent) {
+    if (!(event.target instanceof SVGElement)) {
+      throw new Error("Mouse event's target is not an SVG element");
+    }
```
louisgreiner, line 368: cf. comment about MouseEvent.target check.

louisgreiner (approved):
Good idea to define the type `NodeDragEvent` and similar types, makes the code lighter.

Except a small comment about code refactoring, LGTM, thanks! Tested ✅

src/app/view/editor-main-view/data-views/editor.view.ts
emersion: I've been wondering about this, but (1) three times isn't that bad for two lines of code and (2) I was uneasy about importing `src/app/view/editor-main-view/data-views/d3.utils` from outside of `src/app/view/editor-main-view/`.

louisgreiner: That's fine by me. Let's see if more errors like this are introduced, then we can think on this refactor again. LGTM

Math-R (approved):
Lgtm, thanks for the commit's explanation

src/app/view/editor-main-view/data-views/notes.view.ts
theocrsb, line 425: In the second commit, I get the impression that we can remove the domObj passed in via props

theocrsb (approved):
LGTM for the rest

src/app/view/editor-main-view/data-views/notes.view.ts
emersion: I've kept it for symmetry with the rest of the event handlers, and it's removed anyways in the third commit (superseded by `event`).

PR #1133, opened 2026-06-15
feat(service): emit operation when OrderingAlgorithm changed

louisgreiner:
When changing the ordering algorithm, emit an operation to be able to get this change outside of NGE.


src/app/services/ui/ui.interaction.service.ts
```diff
@@ -326,6 +326,7 @@ export class UiInteractionService implements OnDestroy {
     this.nodeService.transitionsUpdated();
     this.nodeService.connectionsUpdated();
     this.trainrunSectionService.trainrunSectionsUpdated();
+    this.operation.emit(new MetadataOperation({orderingAlgorithm: orderingAlgorithm}));
```
emersion: Nit: `{orderingAlgorithm}` should work here.

emersion (approved):
LGTM!

SharglutDev (approved):
Thank you for the fix !

PR #1139, opened 2026-06-15
feat(service): emit on filterChanged()

louisgreiner:
Emit an operation with the full filter settings payload each time onFilterChanged() is called.

### DIFF
--- a/src/app/models/operation.model.ts
+++ b/src/app/models/operation.model.ts
@@ -3,6 +3,7 @@ import {Trainrun} from "./trainrun.model";
 import {Label} from "./label.model";
 import {Note} from "./note.model";
 import {
+  FilterSettingDto,
   FreeFloatingTextDto,
   LabelDto,
   NodeDto,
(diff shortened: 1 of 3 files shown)


src/app/models/operation.model.ts
```diff
@@ -26,6 +27,7 @@ enum OperationObjectType {
 
 type MetadataDto = {
   trafficSide?: TrafficSide;
+  filterSettingDto?: FilterSettingDto;
```
emersion: It sounds a bit weird to add this to `MetadataDto`, since it's not stored under the `metadata` field in `NetzgrafikDto`.

Maybe a new kind of operation would make more sense?

louisgreiner: Mmh I agree, I tried to stick to the `MetadataDto` definition, but there's another (useless) layer with `FilterDataDto`: https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/blob/lgr/enable-full-name-in-nodes/src/app/data-structures/business.data.structures.ts#L339

A new operation sounds good to me

louisgreiner: I've fixuped this proposition

emersion (approved):
LGTM!

Uriel-Sautron (approved):
LGTM

PR #1142, opened 2026-06-17
fix(view): node name ellipsis depending on connection time

louisgreiner:
To fix two things:
- when no connection time is set
- when connection time is not 1 digit

image

### DIFF
--- a/src/app/view/editor-main-view/data-views/nodes.view.ts
+++ b/src/app/view/editor-main-view/data-views/nodes.view.ts
@@ -513,7 +513,7 @@ export class NodesView {
       .attr(StaticDomTags.NODE_ID, (n: NodeViewObject) => n.node.getId())
       .attr("x", NODE_TEXT_LEFT_SPACING)
       .attr("y", (n: NodeViewObject) => n.node.getNodeHeight() - TEXT_SIZE / 2)
-      .attr("width", "74")
+      .attr("width", (n: NodeViewObject) => this.getNodeLabelTextWidth(n.node))
       .text((n: NodeViewObject) =>
         this.editorView.displayNodesFullName()
           ? n.node.getFullName()
@@ -964,5 +964,15 @@ export class NodesView {
         false,
       );
     }
+  }
+
+  private getNodeLabelTextWidth(node: Node): number {
+    const connectionTime = node.getConnectionTime();
+    let width = 0;
+    if (connectionTime !== null) {
+      width = connectionTime === 0 ? 1 : Math.floor(Math.log10(connectionTime)) + 1;
+    }
+    const connectionTimeTextWidth = width * TEXT_SIZE;
+    return node.getNodeWidth() - connectionTimeTextWidth - NODE_TEXT_LEFT_SPACING;
   }
 }


SharglutDev (approved):
Lgtm and tested

louisgreiner:
Mmmh, I'm definitely not enthousiast of the solution proposed here, but I'm not sure the issue is worth more effort.

src/app/view/editor-main-view/data-views/nodes.view.ts
```diff
@@ -965,4 +965,15 @@ export class NodesView {
...
   }
+
+  private getNodeLabelTextWidth(node: Node): number {
+    const connectionTime = node.getConnectionTime();
+    const connectionTimeTextWidth =
+      (connectionTime === null
+        ? 0
+        : connectionTime < 10
+          ? 1
+          : 0.5 + Math.floor(Math.log10(connectionTime))) * TEXT_SIZE;
```
emersion: Why the special case for `connectionTime < 10`? I would've expected `1 + Math.floor(Math.log10(connectionTime))` to work in all cases.

src/app/view/editor-main-view/data-views/nodes.view.ts
```diff
@@ -965,4 +965,15 @@ export class NodesView {
       );
     }
   }
+
+  private getNodeLabelTextWidth(node: Node): number {
+    const connectionTime = node.getConnectionTime();
+    const connectionTimeTextWidth =
+      (connectionTime === null
+        ? 0
+        : connectionTime < 10
```
emersion: Could we avoid nested ternaries, e.g. with a mutable variable and `if`/`else`?

src/app/view/editor-main-view/data-views/nodes.view.ts
louisgreiner: It is actually the `connectionTime === 0` that is a pain in the ***. This case could simply be reduced to that

src/app/view/editor-main-view/data-views/nodes.view.ts
louisgreiner: would something like that be ok?

emersion: Yup!

src/app/view/editor-main-view/data-views/nodes.view.ts
```diff
@@ -983,4 +965,17 @@ export class NodesView {
       );
     }
   }
+
+  private getNodeLabelTextWidth(node: Node): number {
+    const connectionTime = node.getConnectionTime();
+    let width = 0;
+    if (connectionTime !== null) {
+      width += 1;
+      if (connectionTime >= 10) {
```
emersion: I'm still confused about the special case for 10. Something like `Math.ceil(Math.log10(Math.max(connectionTime, 1)))` would make more sense to me, would that work? (Alternatively, a special case for `connectionTime > 0`.)

src/app/view/editor-main-view/data-views/nodes.view.ts
```diff
@@ -983,4 +965,17 @@ export class NodesView {
...
+        width += Math.floor(Math.log10(connectionTime));
+      }
+    }
+    const connectionTimeTextWidth = width * TEXT_SIZE;
```
emersion, line 975: `TEXT_SIZE` is the character height, not the character width. But oh well, if it's a good enough workaround, let's just leave it at that…

src/app/view/editor-main-view/data-views/nodes.view.ts
louisgreiner: This behaves the same, if the code is ok for you, then by me it's fine.

https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/pull/1142/commits/cbaa0e1291a1bfe2bce3410f0c91555d2a8b8676

emersion (approved):
LGTM!

PR #1143, opened 2026-06-17
feat: ship type definitions in NPM package

emersion:
This change exposes type definitions for NGE data structures in the NPM package. That way, users don't need to duplicate these definitions and get type errors when there's a mismatch.

typeRoots is required to avoid missing minimatch type definitions (because we're stuck with an old version).

Note, operations would be nice to expose as well, but require more thought (currently these pull in a number of internal NGE types, e.g. models).

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
     "start:standalone": "ng serve --configuration=standalone",
     "start:standalonedemo": "ng serve --configuration=standalonedemo",
     "build": "ng run netzgrafik-frontend:ngsscbuild",
-    "build:standalone": "ng build --configuration=standalone",
+    "build:standalone": "ng build --configuration=standalone && tsc --project tsconfig.defs.json",
     "build:standalonedemo": "ng build --configuration=standalonedemo",
     "test": "ng test -c ci",
     "format": "prettier --write .",
@@ -27,6 +27,7 @@
   "files": [
     "dist/*"
   ],
+  "types": "dist/types/types.d.ts",
   "dependencies": {
     "@angular/animations": "^19.2.25",
     "@angular/cdk": "^19.2.19",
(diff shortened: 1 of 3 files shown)


DucNg:
I don't understand how we use this

emersion:
Intended use:

1. Delete this file: https://github.com/OpenRailAssociation/osrd/blob/dev/front/src/applications/operationalStudies/views/Scenario/components/NGE/types.ts
2. Instead, use `import type XXX from '@OpenRail/netzgrafik-editor'`

louisgreiner:
OpenRail package is not available currently. Latest published has been published [a month ago](https://www.npmjs.com/package/@openrail/netzgrafik-editor-frontend?activeTab=versions). 

This cannot be tested for now I guess, but if we merge this in `main`, we can later test it in OSRD's fork](https://github.com/OpenRailAssociation/netzgrafik-editor-frontend) which publishes the [package](https://www.npmjs.com/package/@osrd-project/netzgrafik-frontend?activeTab=versions).

emersion:
The OpenRail package is available and published in CI when a new NGE release is cut. Last NGE release was a month ago.

Of course, we (OSRD) would get this change via our fork.

DucNg (commented):
Works great for `NetzgrafikDto` but it seems like `NGEEvent` is not exported. Do we keep this one on OSRD side?

Image

emersion:
This is something I'd like to fix as well, but it's trickier because `src/app/models/operation.model.ts` imports a number of internal NGE types (e.g. models) that shouldn't be exported as public API.

SharglutDev (approved):
Good idea !

louisgreiner (approved):
Cool!

PR #1144, opened 2026-06-18
fix: orders ports facing ordered neighbors first

### LINKED ISSUE #1140: Port ordering issue with triple parallel paths crossing
### What is the input mode?

Mouse

### What type of issue is it?

Functionality

### What happened?

This ticket is just [an extract](https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/pull/759#issuecomment-3761585618) of a specific bug case from @aiAdrian, from the past review on the new port ordering mode. Most feedbacks were not small enough to surely understand what is an issue and what is not. However, this one clearly is a bug, since the three parallel paths are trivial to unfold.

**Description**

This case shows three absolutely parallel paths (so, a very trivial case to handle), that cross together between two stations at some point, when they clearly shouldn't, when we toggle the "Crossing aware" mode.

**Video**

[PortOrdering_001.webm](https://github.com/user-attachments/assets/a9391f59-f77a-450d-a729-40950d4a0960)

Any number of absolutely parallel paths should never cross with each other.

1. Load [this JSON reticular](https://github.com/user-attachments/files/24902011/netzgrafik.7.json)
2. Toggle the "Crossing aware" mode

Local

Firefox

2.11.1

jacomyal:
This fixes #1140.

There are two separate commits:
- One adds two regression tests (one representing the exact case described in the issue, and another one, simpler, to reduce the problem to something smaller)
- One fixes the code so that the tests pass

The fix is to process sides facing an already-ordered neighbor before others, so a node follows its neighbor's order (case `4a` in the code) instead of locking to its own opposite side (case `2` in the code).

### DIFF
--- a/src/app/services/util/port-ordering.algo.spec.ts
+++ b/src/app/services/util/port-ordering.algo.spec.ts
@@ -1,5 +1,5 @@
 import {reorderNodePorts, optimizePorts} from "./port-ordering.algo";
-import {countCrossingsInNode} from "./port-ordering.crossings";
+import {countAllCrossings, countCrossingsInNode} from "./port-ordering.crossings";
 import {buildNetwork, getTrainrunIDsOnSide} from "./port-ordering.test-helpers";
 
 describe("port-ordering", () => {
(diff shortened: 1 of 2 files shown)


louisgreiner (approved):
LGTM, the added code is so small compared to the added value, thanks. That's so cool!

Also tested on big networks that were formerly mixed-up at places where neighbors were face-to-face but still making section crossing.

image

image

image

(the last case still crosses the grey section in the node, but it avoided more crossings further in the network, so that's very good)

jacomyal:
@louisgreiner Hum, I'm curious with the grey crossings in the last example, it shouldn't happen either, and looks like the same kind of bugs (the crossing should be between two consecutive nodes instead, normally). Could you provide the reticular please?

louisgreiner:
Sure, but the reticular is a bit messy. Here you go (on the right side): 
[reticulaire.json](https://github.com/user-attachments/files/29139151/reticulaire.json)

~~For me, it looks ok~~ edit: that's not ok actually
image

jacomyal:
OK, I observe the same crossing on the previous version (I test here https://openrailassociation.github.io/netzgrafik-editor-frontend/). This is probably another bug somehow. I'll open a new issue.

louisgreiner:
Thanks! Still, this PR fixes many cases (at least in the networks I tried on).

SharglutDev (approved):
Thank you for this quick fix ✅

PR #1146, opened 2026-06-18
feat(view): enable click on tip node

louisgreiner:
https://github.com/user-attachments/assets/27f22049-2f4f-483b-8e3e-b8446fb888b9

### DIFF
--- a/src/app/view/editor-main-view/data-views/trainrunsections.view.ts
+++ b/src/app/view/editor-main-view/data-views/trainrunsections.view.ts
@@ -1688,7 +1688,7 @@ export class TrainrunSectionsView {
         const trainrunSection = d.getExtremitySection(atSource);
         return TrainrunSectionsView.hasWarning(trainrunSection, textElement);
       })
-      .classed(StaticDomTags.TAG_EVENT_DISABLED, true)
+      .classed(StaticDomTags.TAG_EVENT_DISABLED, false)
       .classed(StaticDomTags.TAG_START_TEXT_ANCHOR, (d: TrainrunSectionViewObject) =>
         TrainrunSectionsView.enforceStartTextAnchor(d, atSource),
       )
@@ -1698,7 +1698,12 @@ export class TrainrunSectionsView {
           this.editorView,
           atSource,
         ),
-      );
+      )
+      .on("mouseup", (viewObject: TrainrunSectionViewObject) => {
+        d3.event.stopPropagation();
+        const node = viewObject.getExtremityNode(!atSource);
+        this.editorView.showNodeInformation!(node);
+      });
   }
 
   createNumberOfStopsTextElement(


src/app/view/editor-main-view/data-views/trainrunsections.view.ts
```diff
@@ -1655,6 +1655,7 @@ export class TrainrunSectionsView {
     selectedTrainrun: Trainrun,
     connectedTrainIds: number[],
     atSource: boolean,
+    enableEvents = true,
```
emersion: Is there any spot where this is set to `false`?

src/app/view/editor-main-view/data-views/trainrunsections.view.ts
```diff
@@ -1698,7 +1699,16 @@ export class TrainrunSectionsView {
           this.editorView,
           atSource,
         ),
-      );
+      )
+      .on("mouseup", (viewObject: TrainrunSectionViewObject, i, a) => {
```
emersion: Nit: `i` and `a` seem unused and can be omitted. (Would ease future d3 migration.)

src/app/view/editor-main-view/data-views/trainrunsections.view.ts
```diff
@@ -1698,7 +1699,16 @@ export class TrainrunSectionsView {
...
+        if (enableEvents) {
+          d3.event.stopPropagation();
+          this.editorView.nodeService.setSingleNodeAsSelected(
+            viewObject.getExtremityNode(!atSource).getId(),
+          );
+          this.editorView.uiInteractionService.showNodeBaseData();
```
emersion: Shouldn't we use `this.editorView.showNodeInformation()` as is done elsewhere in the codebase?

src/app/view/editor-main-view/data-views/trainrunsections.view.ts
louisgreiner: Mmh not really, I set it to be aligned with the other methods. It can be removed as it does nothing yet.

src/app/view/editor-main-view/data-views/trainrunsections.view.ts
louisgreiner: Done!

src/app/view/editor-main-view/data-views/trainrunsections.view.ts
louisgreiner: Removed it in the fixup commit

src/app/view/editor-main-view/data-views/trainrunsections.view.ts
louisgreiner: Indeed, I wasn't aware of this method, but since these methods are typed it's way better!

Math-R (approved):
looks good to me

SharglutDev (approved):
Works good ✅ 

But couldn't find what the changes in the scss file and the `.classed` are doing 🤔 Is there for a specific usecase ?

louisgreiner:
This is very surprising, the scss rule fixed a bug that made the line of the trainrun disappear when the mouse's hover the Tip name.

But it seems to have been fixed somehow: the rule is not needed, i'm removing it.

src/app/view/editor-main-view/data-views/trainrunsections.view.ts
```diff
@@ -1698,7 +1698,12 @@ export class TrainrunSectionsView {
           this.editorView,
           atSource,
         ),
-      );
+      )
+      .on("mouseup", (viewObject: TrainrunSectionViewObject) => {
+        d3.event.stopPropagation();
+        const node = viewObject.getExtremityNode(!atSource);
+        this.editorView.showNodeInformation?.(node);
```
emersion: Nit: I'd slightly prefer to use `!` rather than `?` here, a missing callback would indicate that something went wrong in the service setup and needs to be fixed.

emersion (approved):
Looks good to me, feel free to resolve this comment.

PR #1156, opened 2026-06-22
chore(deps): drop explicit @angular-devkit/architect dependency

emersion:
The Angular project template doesn't have this as an explicit dev dependency. It's pulled in by @angular-devkit/build-angular already and we don't use it directly.

Drop it so that Dependabot doesn't bug us with updates.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -33,7 +33,6 @@
         "zone.js": "~0.15.1"
       },
       "devDependencies": {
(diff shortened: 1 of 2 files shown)


Pivouane (approved):
LGTM, thanks for the pr !

Math-R (approved):
lgtm, nice catch, thanks for the clean up

louisgreiner (approved):

PR #1157, opened 2026-06-22
feat: improves port ordering

jacomyal:
This PR should be merged after #1180.

It implements different features to ultimately improve the crossing-aware port-ordering algorithm:

1. There are now two identified types of **clutter**: crossings (already there), and **separations** (two sections that are drawn adjacent at one place but pulled apart at another, eigher within a node or along a link).
2. The candidates generation now tries to address multiple group clutters at once.
3. The optimizer can now accept some crossings within nodes, in order to reduce the crossings between the nodes.
4. The users now can toggle a checkbox, to try moving the clutters from between nodes to within nodes.
5. There is a new `npm run benchmark` task, that runs the optimizer under different weight setups on some sample datasets, and compares the resulting orderings to the alphabetical baseline.

### DIFF
--- a/src/app/services/data/node.service.ts
+++ b/src/app/services/data/node.service.ts
@@ -30,7 +30,7 @@ import {MathUtils} from "../../utils/math";
 import {LabelService} from "./label.service";
 import {FilterService} from "../ui/filter.service";
 import {ConnectionDto, OrderingAlgorithm} from "../../data-structures/technical.data.structures";
-import {optimizePorts} from "../util/port-ordering.algo";
+import {ClutterWeights, optimizePorts} from "../util/port-ordering.algo";
 import {TrainrunSectionValidator} from "../util/trainrunsection.validator";
 import {
   NodeOperation,
@@ -199,8 +199,8 @@ export class NodeService implements OnDestroy {
     });
 
     // Second pass: reorder ports and update routing
-    if (this.currentOrderingAlgorithm === OrderingAlgorithm.CrossingAware) {
-      optimizePorts(this.nodesStore.nodes);
+    if (this.usesOptimizePorts()) {
+      optimizePorts(this.nodesStore.nodes, this.getClutterWeights());
       this.nodesStore.nodes.forEach((node) => {
         node.updateTransitionsRouting();
         node.updateConnectionsRouting();
@@ -1140,8 +1140,30 @@ export class NodeService implements OnDestroy {
     return this.currentOrderingAlgorithm;
   }
 
+  private usesOptimizePorts(): boolean {
+    return this.currentOrderingAlgorithm !== OrderingAlgorithm.Alphabetical;
+  }
+
+  private getClutterWeights(): Partial<ClutterWeights> | undefined {
+    if (!this.usesOptimizePorts()) return undefined;
+
+    const minWeight = 0.2;
+    const pushCrossings =
+      this.currentOrderingAlgorithm === OrderingAlgorithm.ClutterAwarePushCrossings;
+
+    // Boost the separations we keep low: between nodes when pushing crossings in, else within nodes.
+    return {
+      crossingsWithin: minWeight,
+      crossingsBetween: minWeight,
+      separationsWithin: minWeight + (pushCrossings ? 0 : 1),
+      separationsBetween: minWeight + (pushCrossings ? 1 : 0),
+    };
+  }
+
   setOrderingAlgorithm(portOrderingType: OrderingAlgorithm) {
-    this.currentOrderingAlgorithm = portOrderingType;
+    // Any unknown legacy value (the former crossing/separation modes) maps to ClutterAware.
+    this.currentOrderingAlgorithm =
+      portOrderingType in OrderingAlgorithm ? portOrderingType : OrderingAlgorithm.ClutterAware;
   }
 
   getNodes(): Node[] {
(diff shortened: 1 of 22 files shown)


jacomyal:
At the moment, moving the sliders sometimes just stop fixing some trivial conflicts, instead of just allowing to navigate between "trade-offs" situations.

### Example:

This setup is wrong, because the two trainruns going left of XAE/00 should clearly be anchored to its leftest ports:

image

And weirdly, moving the "where?" slider to "within nodes" slightly solves it:

image

jacomyal:
The broken case documented in my previous commit is now solved. I now consider this PR ready for review.

aiAdrian:
I can not see the sliders ! 

What i see ... 
chrome-capture-2026-07-06

jacomyal:
> I can not see the sliders !

Yes indeed, there are no longer sliders, because after many tests they were quite deceptive:

1. The "Where" slider was always changing everything at once at one value precisely, between "all clutters inside the nodes" and "all clutters between the nodes". As it basically acted as a checkbox, I replaced it with one.

2. The "When" slider was also deceptive. Basically, the way the clutters are detected and the candidates are suggested, make it so that crossings are always minimized somehow, while separations were only addressed when the slider was at the maximum "Separations" value. So, I always got better global results when maximizing separations minimization. So I removed the slider.

aiAdrian:
Ok. But I think we should not combine radio-button and check box. We have 3 different cases, please make 3 radio inputs (sbb-selection-content)

src/benchmarks/port-ordering.benchmark.spec.ts
```diff
@@ -0,0 +1,216 @@
+// Benchmarks the port-ordering optimizer's quality: for each dataset and each weight setup, it
```
aiAdrian, line 1: I just wonder why we have just a new `npm run benchmark` which will be manually called. What is the idea behind this and why not using current `npm run test` ?

src/app/view/editor-edit-tools-view-component/editor-edit-tools-view.component.scss
```diff
@@ -79,3 +79,14 @@
   background: var(--sbb-expansion-panel-background-color);
   color: lightgray;
 }
+
+.clutter-options {
+  margin-top: 8px;
+
+  sbb-checkbox {
```
aiAdrian: Please see as well https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/pull/1157/#issuecomment-4893223062

> Ok. But I think we should not combine radio-button and check box. We have 3 different cases, please make 3 radio inputs (sbb-selection-content)

src/app/view/editor-edit-tools-view-component/editor-edit-tools-view.component.html
```diff
@@ -154,6 +154,12 @@ <h2 class="SummaryTitle">{{ "app.view.editor-edit-tools-view-component.edit" | t
         {{ option.name }}
       </sbb-radio-button>
     </sbb-radio-group>
+
+    <div class="clutter-options" *ngIf="showClutterOptions">
+      <sbb-checkbox [(ngModel)]="pushCrossingsIntoNodes" (change)="onClutterOptionChange()">
```
aiAdrian: see as well : https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/pull/1157/#issuecomment-4893223062

> Ok. But I think we should not combine radio-button and check box. We have 3 different cases, please make 3 radio inputs (sbb-selection-content)

aiAdrian (approved):
I approve this -> please just fix the UI -> only radio-button (sbb-selection)

(     )  Alphabetical  
(     )  Crossing aware
(  x  )  Crossing aware ( Push crossings into nodes )

Then i just wonder why we have to split 
(     )  Crossing aware
(  x  )  Crossing aware ( Push crossings into nodes )
into two options

aiAdrian:
I  have approved it -> but please fix radio-buttons (sbb-selection) UI before merging -> then it's fine for me ! Excellent work!

src/benchmarks/port-ordering.benchmark.spec.ts
jacomyal: Basically, I needed a way to clearly evaluate the algorithm quality, to see if some new versions were improving it or not. I used the base `ng test` so that I wouldn't need to mock things up.

But the benchmarks:
- Never fail (they just log the measured quality)
- Are long to run

So adding them into the existing `npm run test` task would bring nothing, and cost some time on each run.

jacomyal:
> Then i just wonder why we have to split ( ) Crossing aware ( x ) Crossing aware ( Push crossings into nodes ) into two options

I think there is no "one better universal" solution between the two options. I think both algos provide different renderings of the trains, and are differently readable. One is clearer within the nodes, and the other is clearer between the nodes.

But if later there is some kind of consensus on one solution being consistently better, we can still merge them then.

src/app/services/data/node.service.ts
```diff
@@ -1140,8 +1140,30 @@ export class NodeService implements OnDestroy {
     return this.currentOrderingAlgorithm;
   }
 
+  private usesOptimizePorts(): boolean {
+    return this.currentOrderingAlgorithm !== OrderingAlgorithm.Alphabetical;
+  }
+
+  private getClutterWeights(): Partial<ClutterWeights> | undefined {
```
louisgreiner: Why do we need the return type to be `Partial` here? We do return all the type's attributes here?

Same question raises around other occurrences of `ClutterWeights` in this PR.

louisgreiner (approved):
PR is surprisingly clear, thanks for the effort. Good idea to add a benchmark and exclude it from the usual tests pipeline.

LGTM, looking forward to use that more and more!

louisgreiner (approved):

src/app/services/data/node.service.ts
jacomyal: Good catch, I just fixed this :+1:

PR #1158, opened 2026-06-23
refactor(view): hunt down "any" types for d3

emersion:
A good chunk of the editor main view logic doesn't use explicit types for d3: either the `any` type is used, either no type is specified at all (ie, `any` is implicitly used). `any` effectively disables TypeScript checks.

Replace all of these with explicit types.

Most of the changes replace `any` with `d3.Selection`. Some of the existing `d3.Selection` types needed adjustments:

- The "datum" type (second generic argument) was undefined, but logic inside functions required a non-undefined type (e.g. `TrainrunSectionViewObject`).
- The "parent" type (third generic argument) was HTMLElement. However, inside an `<svg>` element, the parent may be either an `HTMLElement` (e.g. `<div>`), either an `SVGElement` (e.g. `<g>`). Use the `Element` type instead which covers both. The parent is unused (we never walk up the DOM) but using the correct type is safer.

I'd like to merge this before continuing to work on https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/pull/1123.

### DIFF
--- a/src/app/view/editor-main-view/data-views/connections.view.ts
+++ b/src/app/view/editor-main-view/data-views/connections.view.ts
@@ -11,7 +11,7 @@ import {ConnectionsViewObject} from "./connectionViewObject";
 import {LevelOfDetail} from "../../../services/ui/level.of.detail.service";
 
 export class ConnectionsView {
-  connectionsGroup;
+  connectionsGroup: d3.Selection<SVGElement, undefined, Element, undefined>;
   editorView: EditorView;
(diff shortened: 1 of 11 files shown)


aschantraine (approved):
LGTM

Math-R (approved):
lgtm :) thanks for this one

aiAdrian (approved):
I imaging that this was a very hard work ! Thank you improving this!

PR #1161, opened 2026-06-23
Eliminate some explicit and implicit any

### LINKED ISSUE #1126: Eliminate `any` types

Use proper types instead.

aschantraine:
This PR eliminates some explicit and implicit any in the code base.


emersion (approved):
Thanks a lot for doing this!

PR #1162, opened 2026-06-23
feat: Adding stretching and shrinking of train sections

MoritzWeber0:
In this PR, we've introduced a new algorithm for shrinking and stretching of train sections. You can select specific sections to shrink or stretch. Alternatively there is the option to optimize the whole graph by stretching sections that are below a minimum length or shrink the graph as much as possible.

To stretch or shrink a selection of sections:
1. Select the sections or nodes
2. Press "e" (expand) on the keyboard, for shrinking press "Shift" + "E"

To stretch or shrink the whole graph:
1. Press "i" for expanding the whole graph, for shrinking press "Shift" + "I"

It's especially useful for adding new stops which don't fit into the graph. Instead of moving around all nodes, this automatically stretches the whole graph.

In this PR, there is the assumption of a fixed section size, which doesn't work with long train names. However, there is the follow-up task https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/pull/1166 that solves it and integrates with this PR.

https://github.com/user-attachments/assets/5142d2f2-033d-43e6-b7f2-615fb4fa46d0

### DIFF
--- a/src/app/services/data/node.service.ts
+++ b/src/app/services/data/node.service.ts
@@ -1193,7 +1193,7 @@ export class NodeService implements OnDestroy {
     this.nodesStore.nodes = this.nodesStore.nodes.filter((n) => n.getId() !== nodeId);
   }
 
-  private changeNodePositionWithoutUpdate(
+  changeNodePositionWithoutUpdate(
     nodeId: number,
     newPositionX: number,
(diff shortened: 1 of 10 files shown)


MoritzWeber0:
Superseded by #1169

PR #1163, opened 2026-06-23
fix: orders transition & free-end ports separately

### LINKED ISSUE #1159: Port ordering issue with corner and split-side paths crossing
### What is the input mode?

Mouse

### What type of issue is it?

Functionality

### What happened?

This ticket is an extract of a specific bug case found during the review of #1144, bug that looks like the one from #1140. This feedback is clearly a bug: the crossings below are trivially avoidable. It also is different from the one in #1140, since it occurs both with and without the fix from #1144.

On a single node (NA/BV in the attached reticular), the "Crossing aware" mode leaves two crossings that should not exist:

Image

1. **Two parallel paths** (both grey) entering from the left (LZA/00) and leaving at the bottom (PLN/BV) cross each other in the bottom-left corner. They share both endpoints, so they are absolutely parallel and should never cross. _This bug occurs both before and after the fix at #1144._
2. **Two pass-through** cross near the middle of the node: one path (purple, dashed) goes left-to-right toward the lower-right neighbor (MHV/BV), and another one (grey, dashed) goes to the upper-right neighbor (MPL/BV). They could be very simply reordered within the node to not cross, and that's what the current algorithm aims to do. _This bug only appears after #1144, which explains the different between this screenshot and the one from @louisgreiner in his review on #1144._

The reticular is available [here](https://github.com/user-attachments/files/29139151/reticulaire.json).

There should be no crossing at all within the NA/BV node.

1. Load [reticular.json](https://github.com/user-attachments/files/29139151/reticulaire.json)
2. Toggle the "Crossing aware" mode
3. Look at the NA/BV node

Local

Firefox

2.11.1

jacomyal:
This fixes #1159.

There are three commits:

- One fixes the indentation of the previous regression tests from #1144 
- One adds two regression tests, one per crossing described in the issue #1159, each reduced to a minimal reticular
- One fixes the code so that the tests pass

The issue was that ports without a transition (free-ends) were sorted by the same comparator as transitions, mixing geometry with the tie-break score (case `4b` in the code) into a non-transitive order. That could flip two transitions and make them cross inside the node.

The fix orders transitions by geometry first, then inserts each free-end into that order, so a free end can never flip two transitions.

### DIFF
--- a/src/app/services/util/port-ordering.algo.spec.ts
+++ b/src/app/services/util/port-ordering.algo.spec.ts
@@ -209,32 +209,42 @@ describe("port-ordering", () => {
         expect(groupCrossings).toEqual([]);
       });
     });
-  });
 
-  // Minimal reproduction: two parallel trains turning a 90° corner at HUB,
-  // whose downstream neighbor (MID) is a pass-through continuing to END.
-  // Tested in the 4 corner orientations.
-  describe("Minimal test case", () => {
-    const corners = [
-      {name: "top > right (NE)", arm: {x: 0, y: -100}, mid: {x: 100, y: 0}, end: {x: 200, y: 0}},
-      {name: "right > bottom (SE)", arm: {x: 100, y: 0}, mid: {x: 0, y: 100}, end: {x: 0, y: 200}},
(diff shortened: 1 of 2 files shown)


louisgreiner (approved):

louisgreiner:
Thanks for the exhaustive descripition and comments!

PR #1166, opened 2026-06-23
feat: add distance service to check the length needed between two nodes

bweben:
# Automatic minimum distance calculation
This PR will add a service that checks whether a certain trainrunSection is too small and needs to be expanded by the auto-layouting algorithm.

## How it works
The algorithm looks at each individual trainrunSection and calculates how long the text will be that will be displayed. Meaning the departure time, the train number, arrival time etc.. based on what should be visible. It does not take into account the visibility filter a user could set as then it would need to change the layout when changing the visibility filters. If a trainrunSection is vertical, it will calculate it slightly differently, as the departure and arrival time is still horizontal, meaning the text height instead of the width is used.
As the text is only on the left side of the trainrunSection, it almost doubles the width needed in order to accompany that.

## How it can be changed
In the service, there are some constants that can be changed which change how much the text is taken into account and also which margin is used in addition to the text width.

### DIFF
--- a/src/app/services/util/position.transformation.service.spec.ts
+++ b/src/app/services/util/position.transformation.service.spec.ts
@@ -19,6 +19,7 @@ import {NetzgrafikUnitTesting} from "../../../integration-testing/netzgrafik.uni
 import {ViewportCullService} from "../../services/ui/viewport.cull.service";
 import {PositionTransformationService} from "./position.transformation.service";
 import {Vec2D} from "../../utils/vec2D";
+import {TrainrunDistanceService} from "./trainrun-distance.service";
 
 describe("PositionTransformationService", () => {
   let dataService: DataService;
@@ -103,6 +104,8 @@ describe("PositionTransformationService", () => {
       trainrunSectionService,
     );
 
+    const trainrunDistanceService = new TrainrunDistanceService();
+
     positionTransformationService = new PositionTransformationService(
       trainrunSectionService,
       trainrunService,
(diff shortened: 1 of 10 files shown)


aiAdrian:
The idea originated at the hack4rail.org event in Vienna. It is an excellent concept to calculate the minimal required edge length (trainrunSection). However, I do not recommend using an off‑screen canvas for rendering, because I believe it is (a) computationally expensive and (b) increases complexity. Instead, it should be straightforward to determine the length using the fixed, known text size and the text length (number of characters). The current transformation of what is displayed is not needed, since the rendering engine already handles the projection.

Therefore, the idea should be taken up, discussed, and integrated into the auto.layout.service.ts. Using a dedicated service for this purpose is a very good architectural choice.

=>see https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/pull/1169 -> integrate this approach once the #1169 is merged

bweben:
> The idea originated at the hack4rail.org event in Vienna. It is an excellent concept to calculate the minimal required edge length (trainrunSection). However, I do not recommend using an off‑screen canvas for rendering, because I believe it is (a) computationally expensive and (b) increases complexity. Instead, it should be straightforward to determine the length using the fixed, known text size and the text length (number of characters). The current transformation of what is displayed is not needed, since the rendering engine already handles the projection.
> 
> Therefore, the idea should be taken up, discussed, and integrated into the auto.layout.service.ts. Using a dedicated service for this purpose is a very good architectural choice.
> 
> =>see #1169 -> integrate this approach once the #1169 is merged

One note for using the text length directly: The first implementation was using this method but as the font is not monospaced you need a good heuristic in order to "guess" the width in most cases correctly. That's why I switched to using the OffscreemCanvas, but yes it would be better if this would not be used.

PR #1168, opened 2026-06-24
refactor: add explicit return types for interface methods

emersion:
When the return type is unspecified, it implicitly defaults to "any". Set the return type explicitly to "void" to make sure callers don't use the return value by mistake.


SarahBellaha (approved):

aiAdrian (approved):
Thank you

PR #1170, opened 2026-06-26
fix(services): fix consecutive times when enabling symmetry on non-stop node

### LINKED ISSUE #1141: Non-stops : removing non-symmetry creates invalid consecutive times
### What is the input mode?

Mouse

### What type of issue is it?

Functionality

### What happened?

I start with this network graph : 

[reticulaire (61).json](https://github.com/user-attachments/files/28994354/reticulaire.61.json)

It looks like this

Image

The numbers are : time(consecutiveTime) for : sourceDeparture, travelTime, targetArrival, targetDeparture, backwardTravelTime, sourceArrival

I remove the non-symmetry at B, on both sides, starting from the A side, and then the C side

I now have this : 

Image

The times are weird

I would expect this : 

Image

Use the previous steps

Local

Google Chrome

2.11.1

emersion:
_See individual commits._

Contains the fix, a regression test, and a related simplication refactoring.

Closes: https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/issues/1141

### DIFF
--- a/src/app/services/data/trainrun-section-times.service.spec.ts
+++ b/src/app/services/data/trainrun-section-times.service.spec.ts
@@ -5,6 +5,7 @@ import {TrainrunService} from "../data/trainrun.service";
 import {TrainrunSectionService} from "../data/trainrunsection.service";
 import {BaseDataService} from "../data/basedata.service";
 import {NoteService} from "../data/note.service";
+import {DirectedTrainrunSectionProxy} from "../util/trainrun.iterator";
 import {Node} from "../../models/node.model";
 import {TrainrunSection} from "../../models/trainrunsection.model";
 import {LogService} from "../../logger/log.service";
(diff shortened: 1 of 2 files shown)


src/app/services/data/trainrun-section-times.service.spec.ts
```diff
@@ -484,6 +485,93 @@ describe("TrainrunSectionTimesService", () => {
...
+      .find((port) => port.getTrainrunSectionId() === leftSection.getId())!;
+    const transition = middleNode.getTransitions().find((t) => t.getPortId1() === port.getId())!;
+    nodeService.toggleNonStop(middleNode.getId(), transition.getId());
+
+    // Disable symmetry on both sides of the middle node
+    leftSection.setTargetSymmetry(false);
+    rightSection.setSourceSymmetry(false);
+
+    // Initialize times
+    trainrunSectionService.setTimeStructureToSingleTrainrunSection(leftDirectedSection, {
```
SharglutDev, line 546: Totally not related to your PR but as a nge beginner, I have trouble finding my way around with so many naming for `leftDepartureTime`, `sourceDepartureTime` and `tailDepartureTime`. 
If I see two vertical nodes, left to right direction doesn't help to know what will be the direction.

SharglutDev (approved):
Thanks for the fix, seems to work well

louisgreiner (commented):
Works fine for the consecutive times, however it's sad that the symmetry reset thing works only partially: this works good when both toggles are off, but when having only one off, we are in a buggy state:

Image

| Before  | After (this PR) |
| --- | --- |
| https://github.com/user-attachments/assets/0fe2f0d2-9328-4bba-b179-cd1565ce64f9 | https://github.com/user-attachments/assets/78ee1743-346a-4381-aedb-ea0b089f0a79 |

After discussing with Maëlys, it would be nice to bring back this behavior

src/app/services/data/trainrun-section-times.service.spec.ts
emersion: > Totally not related to your PR but as a nge beginner, I have trouble finding my way around with so many naming for leftDepartureTime, sourceDepartureTime and tailDepartureTime.

Right. We do need all of these variants, but it would be nice to document them in more detail.

`DirectedTrainrunSectionProxy` has some docs explaining why tail/head is useful:

https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/blob/67dd2bece701d3c27f24ca45b41954d74b8cb185/src/app/services/util/trainrun.iterator.ts#L30

> If I see two vertical nodes, left to right direction doesn't help to know what will be the direction.

Left/right is the visual ordering. Top/bottom are equivalent to left/right in this model, and are used for brevity and simplicity. Finding both accurate and palatable names is hard.

emersion:
> however it's sad that the symmetry reset thing works only partially: this works good when both toggles are off, but when having only one off, we are in a buggy state

Good catch! It seems like my refactoring in "refactor(services): simplify keys in updateTrainrunSectionSymmetry()" was a bit too heavy-handed. I've edited that commit to fix it and added another test for this.

louisgreiner (approved):
LGTM! Thanks for re-integrating a "reset symmetry" thing.

PR #1180, opened 2026-06-30
feat: improves port ordering problem decomposition

jacomyal:
This refines how the port-ordering optimizer scopes its work.

Until now, the optimizer ran per connected component (`getConnectedComponents`): a node where two unrelated reticulars merely cross was treated as one block and optimized together, even though ordering one can never affect the other. And since the optimizer uses one single tie-breaker for all crossings, fixing one group crossing was often messing up the rest of the network.

This PR updates the way the optimizer splits the network:

1. The new `getComponents` function groups node sides (not whole nodes) into independent components, with "independent" meaning here that ordering the ports of one can never change the best port ordering of another. Two sides are coupled only when at least two parallel sections or two pass-through transitions tie them. A lone single section is a "bridge" that gets cut. So two components can share a node (each owning a different subset of its ports), or even share a trainrun.

2. `optimizePorts` now runs over these side-components. Each is fed to the optimizer through `getScopedNode`, a read-only `Node` view exposing only that component's ports and transitions, so a shared node has each side optimized by its owning component only, without disturbing the others.

3. `getConnectedComponents` is removed (now dead).

And we end up with more, simpler problems to solve, which strictly improves the global crossings reduction.

_Note to @louisgreiner: I'm quite sure this is the most promising break-through I made on the port ordering algorithm. It's still only optimizing crossings between nodes for now, but I'll update and finalize #1157 soon, to also improve other clutter types._

### DIFF
--- a/src/app/services/util/port-ordering.algo.ts
+++ b/src/app/services/util/port-ordering.algo.ts
@@ -3,7 +3,7 @@ import {Port} from "../../models/port.model";
 import {PortAlignment} from "../../data-structures/technical.data.structures";
 import {Transition} from "../../models/transition.model";
 import {countAllCrossings} from "./port-ordering.crossings";
-import {getConnectedComponents, getPortOppositeNodeId} from "./port-ordering.components";
+import {getComponents, getPortOppositeNodeId} from "./port-ordering.components";
 import {
   ALIGNMENTS_CLOCKWISE_ORDER,
   getOppositeAlignmentScore,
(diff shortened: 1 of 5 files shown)


src/app/services/util/port-ordering.algo.ts
```diff
@@ -3,21 +3,49 @@ import {Port} from "../../models/port.model";
...
+ * The ports stay the same objects, so reordering them still mutates the real network. Everything
+ * else (id, position, ...) is delegated to the real node.
+ */
+function getScopedNode(node: Node, sides: Set<PortAlignment>): Node {
+  const ports = node.getPorts().filter((p) => sides.has(p.getPositionAlignment()));
+  const portIds = new Set(ports.map((p) => p.getId()));
+  const transitions = node
+    .getTransitions()
+    .filter((t) => portIds.has(t.getPortId1()) && portIds.has(t.getPortId2()));
+  const view: Node = Object.create(node);
```
louisgreiner, line 26: Stupid question, why not use the `new Node()` method?

louisgreiner (approved):
Despite the complexity, the comments make it relatively easy to follow, as well as the methods split.

Great idea to split in different hubs to optimize!

Tested, this works very good, as usual, LGTM!

src/app/services/util/port-ordering.algo.ts
jacomyal: We need to have a `view` object, that only exposes a specific subset of the `node`'s ports, but that will still mutate the inner `node`'s data, when its mutation methods are called.

If we create a `new Node()` instance, it won't be tied to the inner `node`'s ports (it will instead carry its very own ports). So, the easiest way I found was to use `Object.create(node)`, that allows only monkey-patching some precise parts of the `node`'s API.

louisgreiner: Ok, I didn't know this trick. Looks great for dealing only with some data transfer. Thanks for the explanations!

PR #1182, opened 2026-07-02
fix(view): travel time position when target is left and section is symmetric

### LINKED ISSUE #1181: Incorrect rendering of traveltimes on certain train segments.
### What is the input mode?

Mouse

### What type of issue is it?

Functionality

### What happened?

Some train segments without asymmetric times and without one‑way configuration display **incorrect traveltimes** in the current version of Netzgrafik-Editor.  
The traveltimes are rendered at the wrong position in the segment, even though the underlying data is correct.  
The times must appear exactly at the same location where they have always been rendered in previous versions.

Screenshot / video:

https://github.com/user-attachments/assets/95d1979a-0a4a-4578-bb28-946f8ede6900

Traveltimes should be rendered at the correct, historically consistent position on each segment, matching the data model and previous versions of Netzgrafik-Editor.

Just open the `standalonedemo` and navigate to `Olten` - `Aarau`

Local, GitHub pages demo

Firefox, Google Chrome, Microsoft Edge, Safari

2.11.1

louisgreiner:
Bug was introduced here: https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/pull/1011/changes

This can be tested using this file: [networkGraphic(14).json](https://github.com/user-attachments/files/29973084/networkGraphic.14.json)

|  Before this PR   | With this PR    |
| --- | --- |
| image |  image  |

Close https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/issues/1181

Resolves https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/pull/1225#discussion_r3664128563


aiAdrian (approved):
LGT!

SharglutDev (commented):
Fix seems to work but don't have the knowledge on vectors or the codebase to challenge it.

Instead of putting in the commit description that it has been generated by AI, could we add a proper description of why it was bugged and why the fix works ?

Maybe also a comment in the code would be nice.

SharglutDev (approved):
Lgtm and tested

src/app/services/util/trainrunsection.routing.ts
```diff
@@ -420,40 +420,61 @@ export class SimpleTrainrunSectionRouter {
...
       ),
     );
 
     const trainrunSectionNumberOfStopsPos = Vec2D.add(
-      Vec2D.scale(Vec2D.add(s1, t1), 0.5),
+      sectionMidPoint,
       nameNumberOfStopsOffsetDirection,
     );
 
+    const nameOffset = Vec2D.sub(trainrunSectionNamePos, sectionMidPoint);
```
emersion: It seems like `nameOffset` is the same as `namePosOffsetDirection`?

```
nameOffset = trainrunSectionNamePos - sectionMidPoint;
nameOffset = (sectionMidPoint + namePosOffsetDirection) - sectionMidPoint;
nameOffset = namePosOffsetDirection;
```

src/app/services/util/trainrunsection.routing.ts
```diff
@@ -420,40 +420,61 @@ export class SimpleTrainrunSectionRouter {
...
+    const travelOffset = Vec2D.sub(trainrunSectionTravelTimePos, sectionMidPoint);
+    const backwardTravelOffset = Vec2D.sub(trainrunSectionBackwardTravelTimePos, sectionMidPoint);
```
emersion: Could we introduce intermediary variables to store offsets before using them to compute `trainrunSectionTravelTimePos`/`trainrunSectionBackwardTravelTimePos`, instead of undoing `Vec2D.add()` after-the-fact?

emersion (commented):
This sometimes switches the meaning of time values above and below the line:

Image

Here "2" should be below, but is displayed above.

louisgreiner:
Sorry again for changing everything in this PR, this one is the final one.

The initial issue was concerning this specific case:
1. when the section is drawn from right-to-left (= target <- source)
2. and when the travel time and backward travel time are equal (then only one is displayed)

The issue was about the trainrun name text element always being above the section line, but the travel times being set on their appropriate side of the line. However, the expected behavior is the following: if only one travel time is displayed, then it should be positioned next to the trainrun name.

This is fixed by this PR, which hasn't been IA generated :oof:

Uriel-Sautron (commented):
Tested; there's still the bug reported by Simon.

louisgreiner:
I reproduced this bug, [reported by Simon](https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/pull/1182#pullrequestreview-4670387307). It is rather about a missing re-render after having modified the time. When manually re-triggering a render, the time element position is correct. See attached video, I'll try to fix it in this PR as well.

https://github.com/user-attachments/assets/ea14d433-1c61-4f0c-97e4-4f0d88f8f21b

I'm a bit surprized because the `TrainrunSectionViewObject.key` should be able to trigger a render, since it listens to all the times...

https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/blob/e2a2e4cd43498d9f2fe34abf2d831051d210e6df/src/app/view/editor-main-view/data-views/trainrunSectionViewObject.ts#L76

louisgreiner:
Bug is fixed by the fixup commit. This is temporary, since in the collapsible nodes branch this cannot happen. I could not manage to add this change at another less-invasive location in the code.

Uriel-Sautron (approved):
LGTM

Math-R (approved):
I tested it with a lot of manipulation, i think it's solved, hope i did not missed anything.
Lgtm !

src/app/services/data/trainrunsection.service.ts
```diff
@@ -961,6 +961,8 @@ export class TrainrunSectionService implements OnDestroy {
   }
 
   trainrunSectionsUpdated() {
+    // This will be removed once path and text considerations are moved to the view layer.
+    this.getSelectedTrainrunSection()?.routeEdgeAndPlaceText();
```
emersion: This update is misplaced IMHO:

- It fires too often, on any trainrun section change.
- It ties updates to selection. When using the perlenkette, the updated trainrun section may not be the selected section.

louisgreiner: Uff, I tried many things but I can't make `trainrun-section-tab` component call `routeEdgeAndPlaceText()` after a time update.

Adapting the `time-stepper` component `(changed)` property does not work either, for some mysterious reason

louisgreiner: In fact, for some odd reason, I cannot make `routeEdgeAndPlaceText()` be called at the very end at another location that this one.

louisgreiner: As a comprise, I moved it in the component, but this is then still buggy. As you can see in the video, it needs another trainrun section update to take this in account and actually recompute the path.

https://github.com/user-attachments/assets/93d11121-2482-4f3b-aa37-b57b6f5ee0e2

I don't think it's worth more effort tbh, since this should be dropped soon. The previous bug is critical, the bug resulting of the compromise is minor I would say.

src/app/services/util/trainrunsection.routing.ts
louisgreiner: I resolve this commit, as the content of this PR has changed.

src/app/services/util/trainrunsection.routing.ts
louisgreiner: I resolve this commit, as the content of this PR has changed.

louisgreiner:
Also resolving https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/pull/1225#discussion_r3664128563

PR #1184, opened 2026-07-02
951 infinity loop on iterator

### LINKED ISSUE #951: Infinity loop on iterator
### What is the input mode?

Mouse

### What type of issue is it?

Functionality

### What happened?

Close https://github.com/OpenRailAssociation/osrd/issues/16252 (issue reported on OSRD)

Importing this file make the browser crash (infinity loop): [networkGraphic.3.json](https://github.com/user-attachments/files/26962579/networkGraphic.3.json)

https://github.com/user-attachments/assets/823f12b0-a9f8-4a19-bacf-7dbda20edb9e

Do not enter in an infinity loop

checkout on `main`, run `npm run start:standalone` then import 

[networkGraphic.3.json](https://github.com/user-attachments/files/26962384/networkGraphic.3.json)

Local

Firefox

2.10.19

aiAdrian:
This PR adds a migration safeguard that removes invalid trainrun sections where source and target nodes are identical before normal DTO migration runs.
It also cleans up related node ports and transitions, and removes trainruns that become orphaned after those invalid sections are deleted.
A localized warning message is emitted when affected sections are found, and the corresponding translations were added to de, en, and fr language files.
A focused unit test was introduced to verify that invalid sections and their dependent references are correctly removed.
Finally, build and targeted tests were executed successfully to confirm that the migration path remains stable and the fix resolves the reported issue.

### DIFF
--- a/src/app/models/node.model.ts
+++ b/src/app/models/node.model.ts
@@ -620,12 +620,13 @@ export class Node {
     return this.getTransitionFromId(transitionId).getIsNonStopTransit();
   }
 
-  removePort(trainrunSection: TrainrunSection) {
-    let portId = trainrunSection.getSourcePortId();
-    if (this.getId() === trainrunSection.getTargetNodeId()) {
-      portId = trainrunSection.getTargetPortId();
+  removePort(trainrunSection: TrainrunSection) { 
+    if (this.getId() === trainrunSection.getSourceNodeId()) {
+      this.ports = this.ports.filter((port) => port.getId() !== trainrunSection.getSourcePortId());
     }
-    this.ports = this.ports.filter((port) => port.getId() !== portId);
+    if (this.getId() === trainrunSection.getTargetNodeId()) { 
+      this.ports = this.ports.filter((port) => port.getId() !== trainrunSection.getTargetPortId());
+    }
   }
 
   removeTransition(trainrunSection: TrainrunSection) {
--- a/src/app/models/node.model.ts
+++ b/src/app/models/node.model.ts
@@ -620,11 +620,11 @@ export class Node {
     return this.getTransitionFromId(transitionId).getIsNonStopTransit();
   }
 
-  removePort(trainrunSection: TrainrunSection) { 
+  removePort(trainrunSection: TrainrunSection) {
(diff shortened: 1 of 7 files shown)


aiAdrian:
This PR resolves the underlying data‑quality issue and prevents the Netzgrafik‑Editor from loading erroneous legacy data. In earlier versions, the Netzgrafik‑Editor had a bug that produced trainrun sections containing only a single node—meaning the source and target were identical. Such invalid sections caused the trainrunIterator to enter an infinite loop. By definition, the Netzgrafik‑Editor cannot contain transitions where the source and target node are the same. This fix ensures that these invalid sections are no longer loaded.

emersion:
This PR mutates loaded JSON files. What I had in mind in https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/issues/951 was to add a check in the piece of code which reaches an infinite loop to throw an error. Even if bad trainrun sections are dropped from the input JSON file, NGE could still generate them by mistake after the file has been loaded.

aiAdrian:
> This PR mutates loaded JSON files. What I had in mind in #951 was to add a check in the piece of code which reaches an infinite loop to throw an error. Even if bad trainrun sections are dropped from the input JSON file, NGE could still generate them by mistake after the file has been loaded.

Yes — I am fully aware of that. But the problem occurs when a user has to load erroneous data from an older version: the version might still be corrupted due to the bug we previously had, and the user may no longer have access to the data at all.
We also cannot fix the data in another service, because we fully rely on the data stored in the JSON, which itself is stored in a blob storage in the backend.
This is why we must perform a migration to ensure that, in such cases, the user does not lose all data.

However, I am also aware that having such a post‑data bug fix is not a very elegant pattern.

But i am open to discuss what kind of pattern we need to implement. Or we decide that this issue will never occur - except we found only once the problem in the reported error - means the reported error is unique.

louisgreiner:
I think this PR can be closed, because https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/pull/1226 has been merged and has fixed the same bug that this PR aimed to.

aiAdrian:
This fix is no longer needed. Thanks to #1226

PR #1200, opened 2026-07-06
refactor(types): export enums as dto types

louisgreiner:
This is required for using these enums externally, as TypeScript does not support supporting enums.


emersion:
For future reference, this has been closed because this doesn't handle number enums correctly and we'd like to avoid hardcoding magic numbers in downstream users.

PR #1203, opened 2026-07-07
test: fix formControl bindings

emersion:
Fixes the following kind of errors when running tests:

    ERROR: 'NG0303: Can't bind to 'formControl' since it isn't a known property of 'input' (used in the 'ProjectsViewComponent' component template).
    1. If 'input' is an Angular component and it has the 'formControl' input, then verify that it is a part of an @NgModule where this component is declared.
    2. To allow any property add 'NO_ERRORS_SCHEMA' to the '@NgModule.schemas' of this component.'

FormsModule and ReactiveFormsModule need to be provided to give access to formControl. For formControl attributes on sbb-ensta components, sbb-ensta modules also need to be imported.


louisgreiner (approved):
Omg, I waited so much for that. Thanks a lot for removing all this noise!

Image

Akctarus (approved):
tested, LGTM!

SarahBellaha (approved):

PR #1205, opened 2026-07-07
Expose number enums and simpler index.html

louisgreiner:
Follow-up of:
- https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/pull/1206

Allows to use the types and number enums externally.

Thanks @Pivouane for the help

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,6 +28,15 @@
     "dist/*"
   ],
   "types": "dist/types/types.d.ts",
+  "exports": {
+    "./index.html": {
+      "default": "./dist/index.html"
+    },
+    ".": {
+      "types": "./dist/types/types.d.ts",
+      "default": "./dist/out-tsc/types.js"
+    }
+  },
   "dependencies": {
     "@angular/animations": "^19.2.25",
     "@angular/cdk": "^19.2.19",
--- a/src/types.ts
+++ b/src/types.ts
@@ -1,5 +1,5 @@
 import type {} from "@angular/localize/init";
 
-export type * from "./app/data-structures/technical.data.structures";
-export type * from "./app/data-structures/business.data.structures";
+export * from "./app/data-structures/technical.data.structures";
+export * from "./app/data-structures/business.data.structures";
 export type * from "./app/models/operation.model";


aiAdrian (commented):
I can not understand the feature in detail @emersion  please review it

src/app/data-structures/operation.data.structures.ts
emersion, line 1: This duplicates operation types. We risk a desync when updating them. I've put together an alternative proposal here: https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/pull/1206

Alternatively, we could perhaps try to add `implements NGETrainrunOperation` to `Operation` to ensure these two copies stay in sync.

package.json
```diff
@@ -28,6 +28,16 @@
     "dist/*"
   ],
   "types": "dist/types/types.d.ts",
+  "exports": {
+    "./dist/index.html": {
+      "default": "./dist/index.html"
+    },
+    ".": {
+      "types": "./dist/types/types.d.ts",
+      "import": "./dist/out-tsc/types.js",
```
emersion: `import` should not be required, because `default` should work for imports. See: https://nodejs.org/api/packages.html#conditional-exports

package.json
```diff
@@ -28,6 +28,16 @@
     "dist/*"
   ],
   "types": "dist/types/types.d.ts",
+  "exports": {
+    "./dist/index.html": {
```
emersion: While at it, can we drop the `dist/` component from the public-facing export path? This would be cleaner.

emersion (commented):
Overall looks good, just two minor comments.

Can the PR title be updated to reflect the narrower scope now that one commit has been dropped?

package.json
louisgreiner: Indeed! Thank you for the documentation.

package.json
louisgreiner: That is unecessary, yes. I removed it in a fixup.

emersion (approved):
Nice, LGTM!

Math-R (approved):
Useful ! thanks

src/types.ts
```diff
@@ -1,5 +1,5 @@
 import type {} from "@angular/localize/init";
 
-export type * from "./app/data-structures/technical.data.structures";
-export type * from "./app/data-structures/business.data.structures";
+export * from "./app/data-structures/technical.data.structures";
+export * from "./app/data-structures/business.data.structures";
```
emersion, line 4: Side note: exporting JS alongside types can be a footgun with classes, see https://github.com/OpenRailAssociation/osrd/pull/17533#discussion_r3571315847

By chance, these files don't contain any class. We should be careful not to export any class in the future.

louisgreiner: I'll try to keep that in mind, many many thanks for the explanations 🙏

PR #1206, opened 2026-07-08
feat: expose operation types to external users

emersion:
This is a follow-up for b49d21771e8e ("feat: ship type definitions in NPM package"). Operation types are received by outside users via event listeners. Exposing them improves type safety and allows external users to drop their NGE type copies.

We don't want to expose constructors because outside users shouldn't create operation objects and we don't want to expose our models (these are an unstable, internal NGE API). Mark constructors as @internal and enable TypeScript's stripInternal option.

@angular/localize/init needs to be imported to make TypeScript happy when it encounters @localize decorators.

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -1,2 +1,5 @@
+import type {} from "@angular/localize/init";
+
 export type * from "./app/data-structures/technical.data.structures";
 export type * from "./app/data-structures/business.data.structures";
+export type * from "./app/models/operation.model";
--- a/src/app/models/operation.model.ts
+++ b/src/app/models/operation.model.ts
@@ -36,6 +36,7 @@ abstract class Operation {
   readonly type: OperationType;
   readonly objectType: OperationObjectType;
 
+  /** @internal */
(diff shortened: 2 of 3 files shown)


src/types.ts
```diff
@@ -1,2 +1,5 @@
+import type {} from "@angular/localize/init";
```
louisgreiner, line 1: I cannot find place where `@Localize` is used in the code. Why is this needed in this PR then?

louisgreiner (approved):
Tested downstream, it works super good! Thanks for helping on this.

I just have a small question, LGTM anyway.

SharglutDev (approved):
Lgtm (not tested)

src/types.ts
emersion: `operation.model.ts` imports various models at the top, and these use `@localize`. TypeScript needs to load all imported files recursively, even if the import ends up being dropped due to `stripInternal` later on.

louisgreiner: Thanks for the explanation!

PR #1207, opened 2026-07-08
feat: add activeFilterSettingId @Input

louisgreiner:
This enables to interact outside of NGE with the activeFilterSetting which is present in the netzgrafikDto.

This enables users to do so: https://github.com/OpenRailAssociation/osrd/pull/17347

### DIFF
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -101,6 +101,20 @@ export class AppComponent implements OnInit {
     this.uiInteractionService.closeTrainrunDialog();
   }
 
+  @Input()
+  get activeFilterSettingId(): number {
+    return this.filterService.getActiveFilterSettingId();
+  }
+
+  set activeFilterSettingId(id: number) {
+    const filterSetting = this.filterService.getFilterSettingFromId(id);
+    if (filterSetting) {
+      this.filterService.setActiveFilterSetting(filterSetting);
+    } else {
+      throw new Error(`FilterSetting with id ${id} not found`);
+    }
+  }
+
   @Output()
   operation: Observable<Operation> = merge(
     this.trainrunService.operation,


aiAdrian (approved):

src/app/app.component.ts
```diff
@@ -101,6 +101,18 @@ export class AppComponent implements OnInit {
     this.uiInteractionService.closeTrainrunDialog();
   }
 
+  @Input()
+  get activeFilterSettingId(): number {
+    return this.filterService.getActiveFilterSettingId();
+  }
+
+  set activeFilterSettingId(id: number) {
+    const filterSetting = this.filterService.getFilterSettingFromId(id);
```
Math-R, line 110: What happens if the id passed to the setter does not match any existing filter setting?

I think we should handle this explicitly with an early return, and maybe log a warning to make the issue easier to debug.

louisgreiner: `getFilterSettingFromId()` will return `null`, then on the next line we do not jump into the `if` condition and instead do nothing. I don't think we should do something else?

https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/blob/3911f385fb611f982d72339d397dd2a584ba2b18/src/app/services/ui/filter.service.ts#L53-L57

emersion: I'd recommend throwing an error. The caller is responsible for picking a valid ID. Silently doing nothing will hide bugs where the caller provides a bogus ID.

emersion (approved):
LGTM with that comment fixed!

src/app/app.component.ts
louisgreiner: Alright, I'm adding this!

Math-R (approved):
lgtm & thanks for the fix

PR #1208, opened 2026-07-10
refactor: make MetadataDto.analyticsSettings optional

louisgreiner:
Also created a constant to avoid magic number.

This makes external users able not to define a specific value and use the natural fallback.


shenriotpro (approved):

emersion (approved):
Sweet, LGTM!

This also better reflects the structure of older JSON files.

PR #1212, opened 2026-07-13
refactor(models): use exhaustive union for Operation

emersion:
Currently, our exported Operation class is open: there is no guarantee that classes in operation.model.ts are the only ones inheriting from our base Operation class. In fact, even NGE users can create their own class inheriting from Operation.

This is cumbersome for users willing to check the type of an operation: checking for type/objectType isn't enough to narrow down the value to a more specific operation class.

To fix this, rename Operation to BaseOperation and don't export it. Instead, export a union of all of our operations. That way, TypeScript is aware that only the listed choices are valid when handling an Operation object.

See discussion here:
https://github.com/OpenRailAssociation/osrd/pull/17533#discussion_r3557315536

### DIFF
--- a/src/app/models/operation.model.ts
+++ b/src/app/models/operation.model.ts
@@ -32,18 +32,18 @@ type MetadataDto = {
   trafficSide?: TrafficSide;
 };
 
-abstract class BaseOperation {
+abstract class BaseOperation<O extends OperationObjectType> {
   readonly type: OperationType;
-  readonly objectType: OperationObjectType;
+  readonly objectType: O;
 
   /** @internal */
-  constructor(type: OperationType, objectType: OperationObjectType) {
+  constructor(type: OperationType, objectType: O) {
     this.type = type;
     this.objectType = objectType;
   }
 }
 
-class TrainrunOperation extends BaseOperation {
+class TrainrunOperation extends BaseOperation<OperationObjectType.trainrun> {
   readonly trainrun: TrainrunDto;
 
   /** @internal */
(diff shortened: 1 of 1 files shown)


louisgreiner (approved):
LGTM, tested in this PR: https://github.com/OpenRailAssociation/osrd/pull/17533#discussion_r3587038152 ✅

emersion:
I've pushed a new commit "refactor(models): make BaseOperation generic on the object type" to address https://github.com/OpenRailAssociation/osrd/pull/17533#discussion_r3587082916

louisgreiner (approved):
Tested the second commit, this works well: https://github.com/OpenRailAssociation/osrd/pull/17533#discussion_r3587299335

Yohh (approved):

src/app/models/operation.model.ts
```diff
@@ -32,18 +32,18 @@ type MetadataDto = {
   trafficSide?: TrafficSide;
 };
 
-abstract class BaseOperation {
+abstract class BaseOperation<O extends OperationObjectType> {
```
Math-R, line 35: nit : i get why you use `O` for the objectType but it can be confusing, as it is not a conventionnal generic like T,K,etc.. wdyt of `TObject` ?

Math-R (approved):
Lgtm, i left you small nit, feel free to resolve

src/app/models/operation.model.ts
emersion: I didn't want to mix up `OperationObjectType` with `OperationType`. TypeScript docs generally use full names instead single letters for generic arguments: if `O` isn't suitable, I think I'd prefer `ObjectType`.

PR #1213, opened 2026-07-13
refactor(service): remove mapLeftAndRightTimes

### LINKED ISSUE #808: Drop `TrainrunSectionHelperService.mapLeftAndRightTimes()`
I believe this is dead code: the `if` inside is never true. Verify this theory and drop it.

louisgreiner:
The if branch was unreachable because the method was only called when nodesOrdered is empty, and, in that case, the two values are compared through getLeftOrTopNode on the same last-non-stop nodes.

Close #808


emersion (approved):

Math-R (approved):
lgtm

PR #1219, opened 2026-07-22
fix: add missing .js file extension to ESM imports

emersion:
ESM requires explicit file extensions:
https://nodejs.org/api/esm.html#mandatory-file-extensions

TypeScript doesn't add file extensions when producing .js files. This results in broken imports when trying to use NGE from Node.

Fix this by adding explicit file extensions on the TypeScript side and enabling rewriteRelativeImportExtensions:
https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-7.html#path-rewriting-for-relative-paths

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -1,5 +1,5 @@
 import type {} from "@angular/localize/init";
 
-export * from "./app/data-structures/technical.data.structures";
-export * from "./app/data-structures/business.data.structures";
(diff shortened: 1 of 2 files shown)


louisgreiner (approved):
Thanks!

SharglutDev (approved):
Lgtm !

PR #1225, opened 2026-07-24
refactor(model): better trainrunSection.model typing and migrate null values to undefined

### LINKED ISSUE #973: Older Netzgrafik cause an warning issue, when missing backward travel times due of wrong warning check

louisgreiner:
_See individual commits_

Some of the methods of `TrainrunSection` were mistyped, mostly because `TimeLockDto.timeFormatter` and `TimeLockDto.warning` can be `undefined`.

This will ease the `strictNullCheck` migration.

Close https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/pull/974
Close https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/issues/973


src/app/view/editor-main-view/data-views/trainrunsections.view.ts
```diff
@@ -177,39 +177,39 @@ export class TrainrunSectionsView {
     switch (textElement) {
       case TrainrunSectionText.SourceDeparture:
         return (
-          trainrunSection.getSourceDepartureWarning().title +
+          trainrunSection.getSourceDepartureWarning()?.title +
```
emersion: It sounds surprising to me to use `?` here. `undefined + ": "` results in the string "undefined: ", which is not a good thing to display to the user.

The `hasWarning()` check above should ensure that these are not undefined, right? In that case, `!` would be more suitable.

We could refactor this to avoid this case, something like:

```ts
const warning = TrainrunSectionsView.getWarning(trainrunSection, textElement);
if (!warning) {
  return "";
}
return warning.title + ": " + warning.description;
```

But it's orthogonal to this PR.

src/app/models/trainrunsection.model.ts
```diff
@@ -572,27 +560,29 @@ export class TrainrunSection {
   }
 
   hasTravelTimeWarning(): boolean {
-    return this.travelTime.warning !== null;
+    return this.travelTime.warning !== null && this.travelTime.warning !== undefined;
```
emersion: Nit: `Boolean(this.travelTime.warning)` would be a shorter way to write this, I think.

emersion:
Would be nice to add a description to the commit message and PR, to explain the motivation for this change.

src/app/models/trainrunsection.model.ts
louisgreiner: I prefer that way too, fixuped it.

louisgreiner:
Sorry for the lack of context, I updated the PR and the commit description.

src/app/view/editor-main-view/data-views/trainrunsections.view.ts
louisgreiner: That's true, I did not test it and I agree the current state is not suitable for the user. I'll add it to the PR.

Pivouane (approved):
LGTM ! Fixups too ! 🦋

src/app/view/editor-main-view/data-views/trainrunsections.view.ts
emersion: It seems like the latest version of this PR still uses `?` around here. Can we switch to `!` to clearly state that we expect these to be non-undefined at this point?

louisgreiner:
Oops, I force pushed on the wrong branch. I made too many changes, I wanted to close this PR and open a new one, close to it but all the commits have changed.

May I ask for a whole re-review from your part? I think the new changes are more valuable.

I've split the PR in different self-explainatory commits.

src/app/view/editor-main-view/data-views/trainrunsections.view.ts
louisgreiner: The current version should be ok now.

Pivouane (approved):
LGTM, thank you for your service ! 🐲

src/integration-testing/test-data/testTransitionNetzgrafik.json
```diff
@@ -343,12 +265,12 @@
           "2": {"x": 980, "y": 572},
           "3": {"x": 1008, "y": 548},
           "4": {"x": 1200, "y": 452},
-          "5": {"x": 1200, "y": 452},
+          "5": {"x": 1200, "y": 476},
```
emersion, line 268: Any idea why some positions have changed in the process?

emersion (approved):
LGTM! Just one question below.

src/integration-testing/test-data/testTransitionNetzgrafik.json
louisgreiner: Edit: because of this bug https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/pull/1182

~Mmh, I honestly do not have the slightest idea about that... What I did to update the JSON files is import then directly after export.~

Since `TrainrunSection.path` is anyway recomputed at each rendering, this is not big of an issue. Also, this property will be moved soon to the view layer.

PR #1226, opened 2026-07-27
fix(services): bail out from propagateConsecutiveTimesForTrainrun() on infinite loop

### LINKED ISSUE #951: Infinity loop on iterator

emersion:
When TrainrunIterator encounters an infinite loop, it stops iteration, leaving some trainrun sections unvisited.

propagateConsecutiveTimesForTrainrun() assumes that all sections of a trainrun will be eventually visited when calling TrainrunIterator repeatedly. This won't be the case when an infinite loop is hit.

Check whether forward progress is made: if an iteration doesn't decrease alltrainrunsections.length, then we've entered an infinite loop. Bail out in this case.

Closes: https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/issues/951

### DIFF
--- a/src/app/services/data/trainrun.service.ts
+++ b/src/app/services/data/trainrun.service.ts
@@ -643,11 +643,20 @@ export class TrainrunService {
       );
 
       // filter all still visited trainrun sections
-      alltrainrunsections = alltrainrunsections.filter(
+      const remainingTrainrunSections = alltrainrunsections.filter(
         (ts) =>
           propDataForward.visitedTrainrunSections.indexOf(ts) === -1 &&
           propDataBackward.visitedTrainrunSections.indexOf(ts) === -1,
       );
+
+      // Ensure forward progress: if we hit an infinite loop in
+      // TrainrunIterator, then we stop iterating leaving some sections
+      // unvisited. If we haven't visited new sections, bail out.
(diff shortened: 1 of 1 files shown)


louisgreiner (approved):
It indeed breaks the infinity loop, nice.

For your information, I tried to then select and delete the trainrun. Two things happened:
- the trainrun is only visually deleted once the trainrun gets unselected
- the node size is not updated after trainrun deletion

I don't think it's worth opening a bug yet.

https://github.com/user-attachments/assets/3a4e0bd8-3ec6-442a-a191-d9e04be5fe48

Math-R (approved):
lgtm

Issue #1167, opened 2026-06-24
Migrate away from save-svg-as-png


emersion:
The project has been archived, it's no longer maintained: https://github.com/exupero/saveSvgAsPng

Find a replacement. We probably don't need a library, see e.g. https://stackoverflow.com/a/74026755/3567989

PR #1233, opened 2026-07-28
Eliminates implicit any to enable `noImplicitAny`

### LINKED ISSUE #1126: Eliminate `any` types

aschantraine:
Replaces all implicit and some explicit `any` types with specific types, simplifying and cleaning up some code around.
Enable `noImplicitAny`.

The typing of `save-svg-as-png` requested the creation of new `.d.ts` file because it is an untyped module and doesn't have a `@types/save-svg-as-png`.
We should [replace this module]( https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/issues/1167) soon anyway.

### DIFF
--- a/src/app/perlenkette/perlenkette.component.ts
+++ b/src/app/perlenkette/perlenkette.component.ts
@@ -275,9 +275,7 @@ export class PerlenketteComponent implements AfterContentChecked, OnDestroy {
   signalHeightChanged(height: number, pathItem: PerlenketteItem) {
     this.perlenketteRenderingElementsHeight.push([pathItem, height]);
     this.renderedElementsHeight = 0;
-    this.perlenketteRenderingElementsHeight.forEach((pItem, idx) => {
-      const el = pItem[0];
-      const height = pItem[1];
+    this.perlenketteRenderingElementsHeight.forEach(([_, height]) => {
       this.renderedElementsHeight += height;
     });
   }
--- a/src/app/perlenkette/perlenkette.component.ts
+++ b/src/app/perlenkette/perlenkette.component.ts
@@ -344,17 +344,12 @@ export class PerlenketteComponent implements AfterContentChecked, OnDestroy {
 
   private getGotoCurrentY(pathItem: PerlenketteItem) {
     let currentY = 0;
-    let next = true;
-    this.perlenketteRenderingElementsHeight.forEach((pItem, idx) => {
-      const el = pItem[0];
-      const height = pItem[1];
+    for (const [el, height] of this.perlenketteRenderingElementsHeight) {
       if (el === pathItem) {
-        next = false;
+        break;
       }
-      if (next) {
-        currentY += height;
-      }
-    });
+      currentY += height;
+    }
     return currentY;
   }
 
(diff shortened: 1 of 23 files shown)


src/index.d.ts
```diff
@@ -0,0 +1,19 @@
+declare module "save-svg-as-png" {
```
louisgreiner, line 1: What is the benefit of moving this typing out of `editor-tools-view.component.ts`? As it is the only component the calls these, I wonder if it makes sense to have this there

louisgreiner (commented):
The rest of the PR LGTM! Tested and noticed no regression.

Thanks a lot for this much needed work!

src/index.d.ts
aschantraine: `save-svg-as-png` is an untyped module and doesn't have a `@types/save-svg-as-png`. In this situation, using a `.d.ts` file is the only solution to type it. If I try to do it inside `editor-tools-view.component.ts`, it becomes a module augmentation which is not working since an untyped module cannot be augmented.
We should [replace this module](https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/issues/1167) soon anyway.

louisgreiner: Alright, let's do that then. Could you also add this to the issue?

aschantraine: Done

src/app/perlenkette/perlenkette.component.ts
```diff
@@ -237,8 +237,8 @@ export class PerlenketteComponent implements AfterContentChecked, OnDestroy {
     return ret;
   }
 
-  getClosestPerlenketteItem(): PerlenketteNode {
-    let retEl: PerlenketteNode = this.perlenketteRenderingElementsHeight.keys()[0];
```
emersion, line 241: Note, would've been nicer to have this in a separate commit, since it's not trivial that it's dead code and it's lost in the other `any`-related changes.

src/app/streckengrafik/services/sg-6-track.service.ts
```diff
@@ -150,9 +152,9 @@ export class Sg6TrackService implements OnDestroy {
   private mergeDistanceCellGridResoultionToBigSegment(
     tracksMatrix: number[],
     nDistanceCells: number,
-  ): any[] {
+  ): number[][] {
```
emersion: Could we use `[number, number, number][]` here instead? This would make it clear that a tuple of numbers is returned, not an array with a variable length.

(Applies to other types down below as well.)

emersion (commented):
Overall looks good! Just one comment (and a note which can be ignored).

src/app/streckengrafik/services/sg-6-track.service.ts
aschantraine: Done

src/app/perlenkette/perlenkette.component.ts
aschantraine: Noted for future PRs !

emersion (approved):
Thanks for working on cleaning up the codebase!

PR #1237, opened 2026-07-29
experimental! feat(view): make editor background rastering grid visible

louisgreiner:
image

The snapping implementation lies on a 32px raster. This commit creates a grid to mimic the same raster as a background layer.

When zooming out below 100%, the grid stroke width is less than 1px. Adding `vector-effect: non-scaling-stroke;` makes it still visible.

This layer should probably only be shown when the user's moving an object. Or make it as an option. This can be done later on.

Some useful information about the different objects' size: https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/blob/main/src/app/view/rastering/definitions.ts#L1

This draft PR is part of an experimental design refinement with @thibautsailly.

### DIFF
--- a/src/app/view/editor-main-view/data-views/editor.view.ts
+++ b/src/app/view/editor-main-view/data-views/editor.view.ts
@@ -45,9 +45,13 @@ import {
 } from "../../../data-structures/business.data.structures";
 import {TrainrunSectionText} from "../../../data-structures/technical.data.structures";
 import {AutoLayoutService} from "../../../services/util/auto-layout.service";
+import {RASTERING_BASIC_GRID_SIZE} from "../../rastering/definitions";
 
 export class EditorView implements SVGMouseControllerObserver {
   static svgName = "graphContainer";
+  private static readonly RASTER_GRID_PATTERN_ID = "editor-raster-grid-pattern";
+  private static readonly RASTER_GRID_DEFS_ID = "editor-raster-grid-defs";
+  private static readonly RASTER_GRID_SPAN = 100000;
   editorMode: EditorMode = EditorMode.NetzgrafikEditing;
   controller: EditorMainViewComponent;
   svgMouseController: SVGMouseController;
@@ -590,6 +594,7 @@ export class EditorView implements SVGMouseControllerObserver {
     this.rootContainer = this.svgMouseController.init(
       this.uiInteractionService.getViewboxProperties(EditorView.svgName),
     );
+    this.createVisibleRasterGrid();
     this.notesView.setGroup(this.rootContainer.append(StaticDomTags.GROUP_SVG));
     this.nodesView.setGroup(this.rootContainer.append(StaticDomTags.GROUP_SVG));
     this.transitionsView.setGroup(this.rootContainer.append(StaticDomTags.GROUP_SVG));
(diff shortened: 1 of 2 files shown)


aiAdrian:
I'm not entirely sure I understand the requirement yet. What problem are we actually trying to solve with this visualization?

I wonder whether displaying the grid will really help users or whether it will just add more visual noise and confusion. If we cannot clearly communicate where the node will actually snap to, the visualization might end up being more confusing than helpful.

Maybe the grid size simply needs to be much larger, for example 128 instead of 32. In that case, the icon on the left could potentially indicate the exact position where the node will be placed after snapping. That would make the target position immediately visible and might eliminate the need for rendering the grid altogether.

Another option could be to make node movement grid-based rather than smooth, so users can directly see where the node will end up while dragging.

image

src/app/view/editor-main-view/data-views/editor.view.ts
```diff
@@ -45,9 +45,13 @@ import {
 } from "../../../data-structures/business.data.structures";
 import {TrainrunSectionText} from "../../../data-structures/technical.data.structures";
 import {AutoLayoutService} from "../../../services/util/auto-layout.service";
+import {RASTERING_BASIC_GRID_SIZE} from "../../rastering/definitions";
```
aiAdrian: NODE_POSITION_BASIC_RASTER => must be used !

aiAdrian:
```
::ng-deep .editor_raster_grid_line {
  stroke: $COLOR_STRECKENGRAPHIK_LINE_GRID_X;
  stroke-width: 1;
  vector-effect: non-scaling-stroke;
  opacity: 0.5; 
  stroke-dasharray: 4, 4;
  shape-rendering: crispEdges;
}

::ng-deep .editor_raster_grid_point {
  stroke: $COLOR_STRECKENGRAPHIK_LINE_GRID_X;
  stroke-width: 1;
  vector-effect: non-scaling-stroke;
  opacity: 0.5;
  shape-rendering: crispEdges;
}
```


``` typescript
   const defs = ownerSvgSelection.append("defs").attr("id", EditorView.RASTER_GRID_DEFS_ID);
    const gridSize = NODE_POSITION_BASIC_RASTER;

    const pattern = defs
      .append("pattern")
      .attr("id", EditorView.RASTER_GRID_PATTERN_ID)
      .attr("patternUnits", "userSpaceOnUse")
      .attr("width", gridSize)
      .attr("height", gridSize)
      .attr("x", 0)
      .attr("y", 0);

    // 1.  
    pattern
          .append("path")
          .attr("class", "editor_raster_grid_line")
          .attr("d", `M ${gridSize} 0 L 0 0 0 ${gridSize}`)
          .attr("fill", "none");

    // 2. 
    pattern
          .append("path")
          .attr("class", "editor_raster_grid_point")
          .attr("d", `M ${(gridSize / 2) - 1} ${(gridSize / 2) - 1} h 2 v 2 h -2 Z`)
          .attr("fill", "none"); 


    this.rootContainer
      .append("rect")
      .attr("class", "editor_raster_grid")
      .attr("x", -EditorView.RASTER_GRID_SPAN)
      .attr("y", -EditorView.RASTER_GRID_SPAN)
      .attr("width", 2 * EditorView.RASTER_GRID_SPAN)
      .attr("height", 2 * EditorView.RASTER_GRID_SPAN)
      .attr("style", `transform: translate(${gridSize / 2}px, ${gridSize / 2}px)`)
      .attr("fill", `url(#${EditorView.RASTER_GRID_PATTERN_ID})`)
```

[chrome-capture-2026-08-24.webm](https://github.com/user-attachments/assets/69f673cd-36cb-4304-95ce-bf29262fbafa)

[poc_001.webm](https://github.com/user-attachments/assets/a0268e1c-39c0-4d37-bdcc-80865f5bd039)

aiAdrian:
I played a little with the idea / this draft branch 

=> see https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/pull/1283


[poc_002.webm](https://github.com/user-attachments/assets/442ef6a5-d04d-4ea2-acfe-d088625c6f01)

PR #1238, opened 2026-07-30
fix(model): make Node.connectionTime nullable

louisgreiner:
Node.connectionTime has always been nullable, since the value of it is set by an input that allows an empty input, resulting in a null value: https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/blob/c7cc792103a7ce04e23e2223172c48ad8092c975/src/app/view/editor-side-view/editor-node-detail-view/editor-node-detail-view.component.html#L36-L42

This PR aligns this behavior and the data model.


aschantraine (approved):
LGTM ! Thanks

SharglutDev (approved):

PR #1246, opened 2026-08-05
Improve junction detection: early‑exit reduces common‑case runtime to…

aiAdrian:
… O(3)

Worst‑case remains O(n), but typical nodes exit after ≤3 ports


emersion (approved):
LGTM!

PR #1247, opened 2026-08-05
fix: persist scaled graph positions

tomatotomata:
## Summary

The Ctrl+mousewheel scaling path updates node and note models in memory, but it does not emit the update operations that the host uses to persist edits. After a refresh, scaled positions therefore return to their previous values.

This patch emits the same `update` operations used by the alignment actions for every node and note changed by both full-graph and selected-element scaling. The added specs cover full-graph and multi-selected-node scaling.

## Validation

- `npx ng test -c ci --watch=false --include='src/app/services/util/position.transformation.service.spec.ts'`
- `npx tsc --noEmit --project tsconfig.spec.json`
- `git diff --check`

The focused Angular run passed all 6 specs in ChromeHeadless.

Related to OpenRailAssociation/osrd#18002

### DIFF
--- a/src/app/services/util/position.transformation.service.spec.ts
+++ b/src/app/services/util/position.transformation.service.spec.ts
@@ -17,7 +17,6 @@ import {LoadPerlenketteService} from "../../perlenkette/service/load-perlenkette
 import {NetzgrafikUnitTesting} from "../../../integration-testing/netzgrafik.unit.testing";
 import {ViewportCullService} from "../../services/ui/viewport.cull.service";
 import {PositionTransformationService} from "./position.transformation.service";
-import {Operation} from "../../models/operation.model";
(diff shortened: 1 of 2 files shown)


aiAdrian (approved):
LGTM - but i can not test it due i don't have access to the application using this.

louisgreiner (changes requested):
I’m not sure the tests are needed. It adds a maintenance cost without adding big value on the stability of the app. I’d rather remove them. Rest of code looks ok but i did not test.

Also, I think it would be better if automated PR proposed by AI were tested before being approved by a maintainer.

DucNg (changes requested):
Tested in OSRD using npm link, it does not seem to fix the issue, it probably requires some changes on OSRD side as well.

PR #1248, opened 2026-08-05
Improve TrainrunSection lookup performance with a trainrunId index map

aiAdrian:
This PR adds an in‑memory index for fast TrainrunSection lookup by trainrunId. It introduces a Map<number, TrainrunSection[]>, keeps it synchronized on load/add/remove, and updates the service to use direct map access instead of full‑list filtering. This significantly improves lookup performance, especially when iterating all sections of a trainrun by ID.


src/app/models/trainrunsection.model.ts
```diff
@@ -248,6 +248,9 @@ export class TrainrunSection {
   }
 
   setTrainrun(trainrun: Trainrun) {
+    // Don't use this methode - you should use the TrainrunSectionService.updateTrainrunReference() method instead,
```
louisgreiner: I'm worried about the needed removal of `TrainrunSection.setTrainrun()` uses. Should we set the method's scope as `private` then, to prevent its use?

Or, set `trainrun: Trainrun` as optional parameter of the constructor? To remove any call of `setTrainrun()` outside of the class?

I'm not sure how to proceed or if it's a good idea.

src/app/models/trainrunsection.model.ts
```diff
@@ -248,6 +248,9 @@ export class TrainrunSection {
   }
 
   setTrainrun(trainrun: Trainrun) {
```
louisgreiner: This method is still called at these places:
- `TrainrunSectionService.createTrainrunSection()`
- `TrainrunSectionService.copyTrainrunSection()`
- `TrainrunSectionService.createNewTrainrunSectionFromDto()`

These need to call `TrainrunSectionService.updateTrainrunReference()` instead.

src/app/services/data/trainrunsection.service.ts
```diff
@@ -133,6 +135,43 @@ export class TrainrunSectionService implements OnDestroy {
...
   }
 
+  private rebuildSectionIndex() {
+    this.sectionLookupByTrainrunId.clear();
+    for (const section of this.trainrunSectionsStore.trainrunSections) {
+      const id = section.getTrainrunId();
+      if (!this.sectionLookupByTrainrunId.has(id)) {
+        this.sectionLookupByTrainrunId.set(id, []);
+      }
+      this.sectionLookupByTrainrunId.get(id)!.push(section);
```
louisgreiner: This is the exact same of that is used in `addSectionToIndex()`. This could be moved out to another method, called by both `rebuildSectionIndex()` and `addSectionToIndex()`.

src/app/services/data/trainrunsection.service.ts
```diff
@@ -133,6 +135,43 @@ export class TrainrunSectionService implements OnDestroy {
...
+    this.trainrunSectionsStore.trainrunSections.push(trainrunSection);
+    this.addSectionToIndex(trainrunSection);
+  }
+
+  private removeSection(trainrunSection: TrainrunSection) {
+    this.trainrunSectionsStore.trainrunSections =
+      this.trainrunSectionsStore.trainrunSections.filter(
+        (e) => e.getId() !== trainrunSection.getId(),
+      );
+    this.rebuildSectionIndex();
```
louisgreiner, line 161: This could be done with the lookup table as well I think. But perhaps it's not worth the complexity. A comment about that might be nice.

(feel free to ignore)

src/app/services/util/port-ordering.test-helpers.ts
```diff
@@ -46,6 +46,8 @@ export function buildNetwork(def: {
       const ts = new TrainrunSection();
       ts.setSourceNode(sourceNode);
       ts.setTargetNode(targetNode);
+      // just update the trainrun reference in this helper function, because this is only used in test and we dont
+      // want to use the fully trainrunsectuion functionality here, which would require a lot of other services to be initialized
```
louisgreiner: This might be removed if we move the `setTrainrun()` to the constructor, as mentionned above.

louisgreiner (commented):
This looks very faster indeed. This requires a big zoomed out graph to test it: [reticulaire(11).json](https://github.com/user-attachments/files/30783287/reticulaire.11.json) (Performance analyzer from browser makes it clear too).

The main thing for me is that the `Trainrun.setTrainrun()` method is very dangerous now and if used at wrong places, will create bugs. It needs to be prevented.

src/app/services/data/trainrunsection.service.ts
aiAdrian: Yes it could - but the rebuild might more easy and fast ...

emersion (commented):
I would like to suggest a simpler, alternative proposal which retains a single source of truth: https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/pull/1306

src/app/models/trainrunsection.model.ts
aiAdrian: Private doesn't work - typescript doesn't have the protected concept implemented.

aiAdrian: Private doesn't work - typescript doesn't have the protected concept implemented.

emersion:
> Friend‑Token / Object Capability Pattern — Design Rationale

I would rather leave out this part. It adds a lot of noise and complexity for little benefit IMHO.

aiAdrian:
@emersion - Do I understand you correctly? You’re suggesting that we can remove the token because setTrainrun() don't need protection. That would also be fine for me.

PR #1264, opened 2026-08-12
Feature - Quicksearch goto navigation for nodes and trainruns

### LINKED ISSUE #803: Quick‑Search & Goto Navigation for Nodes and Trainruns
### Estimated Size (weeks)

1

### Quick‑Search & Filtering Concept
The search bar idea is very good. I would separate the quick‑search functionality from the filtering feature. For the quick search, we should reuse the same behavior we already have in the Pearls View:

- Clicking a node name triggers a goto action and centers the node in the viewport (camera moves).
- Clicking a trainrun section in the Pearls View centers it.
- Clicking a trainrun section in the Netzgrafik scrolls the Pearls View so that the corresponding section is centered.

#### Quick‑Search for a node (Version 1)
- User selects a node
- User presses a Goto button Image  next to it
- View centers on that node
 
Image

#### Quick‑Search for a trainrun  (Version 2)
- User selects a trainrun
- User presses Goto button  Image  
- System jumps to the start `top-left` node. Second click on the search button the camera moves to the `bottom-right` end node of the trainrun section. 
- If multiple trainruns with given cat + name exists, it cyclic iterates from one to the other element, allways start node - end node, ... 

Image

#### Allowed input:
 
| text     |  data source | example      |
| ------- | ---| ---------- | 
| Node    |  node.betriebspunktName | LTH      |
| Node |  note.fullName | Paris | 
| Trainrun |  trainrun.cat.shortName + ` ` + trainrun.name | IC 25  |

- Autocomplete -> based on node or trainrun name 
- If a trainrun name or node name appears multiple times, the name is supplemented with a `*`. Each click on the search button( Image  ) moves to the next occurrence of that name (cyclic iteration) and calls the goto. This enables a complete search, but also fast duplicate finding for given trainrun name or node name.
- The input field does only allow zero or one element: trainrun or node reference 
- Press on the chip element (x) the element gets deleted
- Click into the input field and press `back` or `delete` the elements gets deleted
- When the node/trainrun gets deleted, renamed, .. -> no longer exits, the elements gets as well removed from input field or if only one occurance still exits the `*` gets automatically removed

#### Input Behavior
Empty input:  
- Shows the default placeholder text: `enter node / trainrun...`.
- Goto button is disabled 

Value present:  
- When a node or trainrun name is present, it remains visible as long as the chip input (single selection) is not deleted or overwritten. The user can click multiple times on the `goto` button( Image  ). 

Go‑to button ( Image  ):
- Enabled when a valid node/trainrun is selected
- Disabled otherwise

#### Open Question
- Should the entered data be written into the netzgrafik.json for persistence, or should it remain only for the duration of the current session?
- Should the trainrun (goto) be automatically selected just after camera moving, when click goto button? 

### Mock-ups

Image

### Acceptance Criteria

- [ ] Menu bar has a  new quick search entry point (input field + search button) for navigate to found element { trainrun end or start node | node }
- [ ] User can select a trainrun or a node in the quick search input field with the help of autocomplete 
- [ ] Goto works like described in the description 

### Implementation Plan

- Add UI element with autocomplete
- Connect the goto function 

### Tests

UI has to be manually tested

### Definition of Ready

- [x] Mock-ups are complete and validated (PO/UX-UI)
- [ ] ACs are clear and have been reviewed by another refiner (PO/UX-UI)
- [ ] Implementation plan has been written and validated by another maintainer (Technical)
- [x] Validated by Adrian (General)

aiAdrian:
This new feature allows user to navigate very quickly throught the Netzgrafik-Editor. The use can either search for a node or for a trainrun. 

- New left-side panel : Search added 
- Search-Component (main) added (container) 
- Search Component for Node search added 
- Search Component for Trainrun search added (and re-used the node from pearls view)

[Navigator_001.webm](https://github.com/user-attachments/assets/aae427df-d690-4153-8272-5ffa6a111b06)

[Navigator_002.webm](https://github.com/user-attachments/assets/b11a5181-df2a-4f05-8692-8a1197e994f9)


src/app/perlenkette/service/load-perlenkette.service.ts
```diff
@@ -75,6 +81,40 @@ export class LoadPerlenketteService implements OnDestroy {
     return this.perlenketteTrainrun$;
   }
 
+  public getOrderedNodeEntriesForTrainrun(trainrun: Trainrun): OrderedTrainrunNodeEntry[] {
+    const allItems = this.getPerlenketteItem(trainrun);
+    const entries: OrderedTrainrunNodeEntry[] = [];
+
+    allItems.forEach((item, index) => {
+      if (item.isPerlenketteNode()) {
+        if (!item.isPerlenketteNode()) {
```
louisgreiner: Hmm the second if will never be reached

louisgreiner (commented):
This feature is indeed very useful! I'll show this to the business next week or the week after (vacation).

However, my biggest point of concern is that the PR is massive and the code seems very different than the rest of the code base. It will surely take some time to review.

As for a more personal point of view, the classes `PerlenketteNode` and `PerlenketteSection` are difficult to work with and all the code that uses that is difficult to maintain. An example of that is the code added in `getOrderedNodeEntriesForTrainrun()`, which is quite scary 😄

If some more comments were added, it would be easier to dive in it. (what is a "gap"? `hasGapAfter` i.e.)

src/app/perlenkette/service/load-perlenkette.service.ts
aiAdrian: uch - i will check this. very bad copy paste issue!

aiAdrian: resolved - thank you for spotting this out

PR #1270, opened 2026-08-17
ci: add Docker build check

emersion:
Check that the Docker image builds correctly when merging a pull request.

Previously, the Dockerfile was only checked when publishing a release. This patch ensures that we don't hit a build failure when performing a release.


louisgreiner (approved):

aiAdrian (approved):
LGTM!

PR #1271, opened 2026-08-18
feat(service): emit operations after scale netzgrafik

louisgreiner:
Superseeds https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/pull/1247 (as it's been opened by a bot)

Tested on a OSRD/NGE environment, this works.

### DIFF
--- a/src/app/services/util/position.transformation.service.ts
+++ b/src/app/services/util/position.transformation.service.ts
@@ -6,7 +6,12 @@ import {NoteService} from "../data/note.service";
 import {Vec2D} from "../../utils/vec2D";
 import {Node} from "../../models/node.model";
 import {ViewportCullService} from "../ui/viewport.cull.service";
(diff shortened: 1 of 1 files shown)


emersion (approved):

DucNg (commented):
The feature works but makes many request to the backend. Maybe we could add a debounce?

It makes 2 requests per node for each scale steps.

https://github.com/user-attachments/assets/886ba472-603e-476a-9681-eca20abbe010

DucNg (approved):
I prefer the debounce version even if it adds a little more complexity. We can reuse this logic for other part of the code if we want to but it's concise enough to only use here for now.

emersion (commented):
It makes sense to me to only emit a single event per node per scaling step, but it doesn't make sense to add a 1s debounce. Some use-cases for standalone mode require up-to-date information without a delay, it should be up to the caller to debounce all updates if they need to.

PR #1292, opened 2026-08-25
fix(perlenkette): render and scroll correctly in Safari

### LINKED ISSUE #1026: OS/X - Pearlsview - Line view not properly shown in Safari
### What is the input mode?

Mouse

### What type of issue is it?

Functionality, Accessibility

### What happened?

Image

https://github.com/user-attachments/assets/d7860336-c6d3-4ff7-9bdc-566948d16d82

The stations are not aligned with the line and do not scroll up/down with the line when using the scrolling wheel.

_No response_

1. Open the Netzgrafik Editor in Safari.
2. Double click on a line

Flatland

Safari

2.10.9

simonbucher:
Fixes #1026

## Problem

When a trainrun is selected, the perlenkette is rendered torn apart in Safari and the nodes
(Betriebspunkte) are not aligned with the line and do not scroll along, while everything looks fine
in Chrome.

## Cause

The content was rendered inside a `<foreignObject>` and scrolled by moving the `viewBox` of the
wrapping `<svg>`. WebKit does not apply that viewBox transformation reliably to positioned html
content inside a `foreignObject`. The nodes use `position: relative` and therefore stayed in place
while the sections scrolled.

## Solution

- Removed the `<svg>` / `<foreignObject>` wrapper, the content is now plain html inside a viewport
  with `overflow: hidden`
- Scrolling is done with a css `transform`; the former vertical centering is kept, so the rendering
  in Chrome is unchanged
- The drag delta is derived from `clientY` instead of `MouseEvent.movementY`, because WebKit reports
  screen pixels there, which made dragging scroll twice as fast on retina displays

## Test

Manually verified in Safari and Chrome: select a trainrun, scroll with the mouse wheel and by
dragging.

### DIFF
--- a/src/app/perlenkette/perlenkette.component.html
+++ b/src/app/perlenkette/perlenkette.component.html
@@ -168,50 +168,42 @@
     <div class="sepdiv"></div>
   </div>
 
-  <div class="svg-container">
-    <svg
-      #svgPerlenkette
-      [attr.height]="contentHeight * 1.25"
-      [attr.width]="contentWidth"
-      (mousemove)="changeSvgMousePosition($event)"
-      [attr.viewBox]="getPerlenketteViewBox()"
+  <div
+    class="svg-container"
+    #svgPerlenkette
+    [style.height.px]="getViewportHeight()"
+    (mousemove)="changeSvgMousePosition($event)"
+  >
+    <div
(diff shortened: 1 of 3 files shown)


aiAdrian (approved):
LGTM!

PR #1293, opened 2026-08-25
Move Connection.path to ConnectionsViewObject

### LINKED ISSUE #1223: Move `Connection.path` to `ConnectionViewObject.path`
`Connection.path` is used only by the view layer, so it should be there and not in the model layer.

Move it to `ConnectionViewObject.path`, as it has been done for `TrainrunSection.path` in this PR for `TrainrunSectionViewObject`: https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/blob/mrd/new-nodes-management-system/src/app/view/editor-main-view/data-views/trainrunSectionViewObject.ts

YatoVoid:
Closes #1223.

`Connection.path` was only ever read by the view layer (`connections.view.ts`, `connectionViewObject.ts`), so it belongs on `ConnectionsViewObject`, the same move already done for `TrainrunSection.path` → `TrainrunSectionViewObject.path`.

`Node.computeConnectionRouting()` is removed along with it. The model no longer computes or stores connection routing; `ConnectionsViewObject` now computes it once in its constructor via `SimpleTrainrunSectionRouter.routeConnection()` and exposes it as `path`. `Node.addConnectionAndComputeRouting()` and `Node.updateConnectionsRouting()` keep their `ConnectionValidator.validateConnection()` calls, which don't depend on the path.

`ConnectionsView.displayConnections()` used to viewport-cull raw `Connection`s via `getPath()` before building view objects. Since the path now only exists once a view object is built, the cull check moved into `createTransitionViewObjects()`, run against the already-computed `path` on each view object right after construction, instead of upfront on the raw connection list. No other behavioral change.

Updated the one test that read `Connection.getPath()` to compute the same routing directly via `SimpleTrainrunSectionRouter.routeConnection()`.

Verified:
- `npx tsc --noEmit`: same pre-existing unrelated errors as a clean `main` checkout (stale `@types/glob`/`minimatch` types, `src/types.ts` `.ts`-extension imports), nothing new.
- `ng lint`: clean on the changed files.
- `ng test -c ci`: full suite, 635 of 635 passing.
- `ng build --configuration=standalone`: clean.

### DIFF
--- a/src/app/models/connection.model.ts
+++ b/src/app/models/connection.model.ts
@@ -1,5 +1,4 @@
 import {ConnectionDto, WarningDto} from "../data-structures/technical.data.structures";
-import {Vec2D} from "../utils/vec2D";
 
 export class Connection {
   private static currentId = 0;
@@ -8,7 +7,6 @@ export class Connection {
   private port1Id: number;
   private port2Id: number;
 
-  private path: Vec2D[];
   private warning: WarningDto = null;
 
   private isSelected = false;
@@ -52,14 +50,6 @@ export class Connection {
 
   getPortId2(): number {
     return this.port2Id;
-  }
-
-  setPath(path: Vec2D[]) {
-    this.path = path;
-  }
-
-  getPath(): Vec2D[] {
-    return this.path;
   }
 
   setWarning(warningTitle: string, warningDescription = "") {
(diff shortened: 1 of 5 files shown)


louisgreiner (approved):
Great thanks! LGTM and tested.

`Transition.path` would need the same refactor: https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/issues/1299

aiAdrian (approved):
LGTM! Thank you

PR #1295, opened 2026-08-26
Implement iteration protocol in TrainrunIterator

hhirtz:
Allows writing `for (const pair of iterator){ ... }` instead of `while (iterator.hasNext()) {...}`

the first 2 commits fix strictNullChecks from the file, the third commit does the thing and the last commit changes while loops to for loops.

There are a bunch of while loops in `trainrun.service.ts` that are meant to find the last pair of the iterator, i haven't changed those. Maybe it's better to have a `ìterator.last()` method for this use case.

Closes #883


emersion (approved):
Nice, LGTM!

Thanks for splitting the work in digestible chunks :)

PR #1297, opened 2026-08-26
refactor(model): lower symmetry to trainrun

louisgreiner:
This aims to resolve part of the performance issue addressed here: https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/pull/1248#issuecomment-5206068911

`sourceSymmetry` and `targetSymmetry` are two attributes of a `TrainrunSection` which composes a `Trainrun`. Checking if a `Trainrun` is symmetric implies to iterate over all of its `TrainrunSection`, which can be highly consuming.

This PR brings the symmetry to the `Trainrun` model, allowing to check only one field when one wants to check the symmetry of it.

I'm not sure this is a good pratice, because nothing really avoids the new `Trainrun.symmetric` field to be coherent with the `TrainrunSection.{source/target}Symmetry` of the trainrun.

### DIFF
--- a/src/app/models/trainrun.model.ts
+++ b/src/app/models/trainrun.model.ts
@@ -26,6 +26,8 @@ export class Trainrun {
   private labelIds: number[];
   private direction: Direction;
 
+  private symmetric: boolean = true;
+
   constructor(
     {
       id,
@@ -175,5 +177,13 @@ export class Trainrun {
       labelIds: this.labelIds,
       direction: this.direction,
     };
+  }
+
+  isSymmetric(): boolean {
+    return this.symmetric;
+  }
+
+  setSymmetric(isSymmetric: boolean): void {
+    this.symmetric = isSymmetric;
   }
 }
--- a/src/app/models/trainrunsection.model.ts
+++ b/src/app/models/trainrunsection.model.ts
@@ -254,10 +254,16 @@ export class TrainrunSection {
 
   setSourceSymmetry(sourceSymmetry: boolean) {
     this.sourceSymmetry = sourceSymmetry;
+    if (!sourceSymmetry) {
+      this.trainrun.setSymmetric(false);
+    }
   }
 
   setTargetSymmetry(targetSymmetry: boolean) {
     this.targetSymmetry = targetSymmetry;
+    if (!targetSymmetry) {
+      this.trainrun.setSymmetric(false);
+    }
   }
 
   resetSymmetry() {
--- a/src/app/models/trainrunsection.model.ts
+++ b/src/app/models/trainrunsection.model.ts
@@ -266,14 +266,10 @@ export class TrainrunSection {
     }
   }
 
-  resetSymmetry() {
-    this.sourceSymmetry = true;
-    this.targetSymmetry = true;
-  }
-
   setAsymmetry() {
     this.sourceSymmetry = false;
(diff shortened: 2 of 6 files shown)


src/app/models/trainrunsection.model.ts
```diff
@@ -254,10 +254,16 @@ export class TrainrunSection {
...
+    if (!sourceSymmetry) {
+      this.trainrun.setSymmetric(false);
+    }
   }
 
   setTargetSymmetry(targetSymmetry: boolean) {
     this.targetSymmetry = targetSymmetry;
+    if (!targetSymmetry) {
+      this.trainrun.setSymmetric(false);
+    }
```
hhirtz, line 266: Shouldn't this set the trainrun to be symmetric if all sections are symmetric here?

I don't really understand how it can work even though setting `setSourceSymmetry` to true wouldn't set the trainrun's symmetry back to true

louisgreiner: True, fixed it.

src/app/services/data/trainrun-section-times.service.ts
```diff
@@ -595,6 +595,9 @@ export class TrainrunSectionTimesService {
     const {leftSymmetry, rightSymmetry} = this.symmetryStructure;
     leftSection.setTailSymmetry(leftSymmetry);
     rightSection.setHeadSymmetry(rightSymmetry);
+    if (this.areAllSectionsSymmetric()) {
+      this.selectedTrainrunSection.getTrainrun().setSymmetric(true);
+    }
```
hhirtz: No need to go through all sections if we're setting symmetry to false:

```suggestion
    if (leftSymmetry && rightSymmetry && this.areAllSectionsSymmetric()) {
      this.selectedTrainrunSection.getTrainrun().setSymmetric(true);
    }
```

But also these lines should probably go into trainrunsection's setSourceSymmetry and setTargetSymmetry, like what's done for the case where we set the symmetry to false.

src/app/models/trainrun.model.ts
```diff
@@ -26,6 +26,8 @@ export class Trainrun {
   private labelIds: number[];
   private direction: Direction;
 
+  private symmetric: boolean = true;
```
hhirtz, line 29: why not

```suggestion
  public symmetric: boolean = true;
```

hhirtz (commented):
Looks good to me aside from the first comment 👍 

This loop can also be removed i think:
https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/blob/71b837ddc9fd5b2981176f5ff99629c2db2588f2/src/app/view/dialogs/trainrun-and-section-dialog/trainrunsection-tab/trainrun-section-tab.component.ts#L100-L104

src/app/models/trainrun.model.ts
louisgreiner: Model class attributes are read and modified through the dedicated getter and setter. I guess it's to protect this attribute from unwanted modification.

src/app/services/data/trainrun-section-times.service.ts
louisgreiner: True! Added it in `updateTrainrunSectionSymmetry`.

However, it's already there in `setSourceSymmetry` and `setTargetSymmetry`.

PR #1298, opened 2026-08-26
chore(typescript): enable `strict` mode without `strict{NullChecks,FunctionTypes}`

emersion:
The `strict` option enables a range of checks: https://www.typescriptlang.org/tsconfig/#strict

Enable everything except `strictNullChecks` and `strictFunctionTypes` because these trigger a few hundred errors. We should fix these two incrementally.

References: https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/issues/1289

PR #1301, opened 2026-08-26
add more files to the strictNullChecks config

hhirtz:
I don't know if it's something we want, but adding strictPropertyInitialization only issued a warning for the logger (which ended up being about an unused property instead? i don't see where `location` is used), vec2D and the auth service.

Also went through folders in src/app and added them to the strictNullChecks config.

### DIFF
--- a/src/app/services/auth/auth.service.ts
+++ b/src/app/services/auth/auth.service.ts
@@ -11,7 +11,7 @@ import {environment} from "../../../environments/environment";
 export class AuthService {
   // Promise that resolves once the login has been successful.
   // This only works for forceful logins.
-  readonly initialized: Promise<unknown>;
+  readonly initialized: Promise<unknown> | undefined;
 
   get claims(): {email: string; name: string; roles: string[]} {
     return this.oauthService.getIdentityClaims() as {
@@ -31,6 +31,7 @@ export class AuthService {
     location: Location,
   ) {
     if (environment.disableBackend) return;
+    if (!environment.authConfig) throw Error("missing authConfig from environment");
     this.oauthService.configure(environment.authConfig);
     this.oauthService.setupAutomaticSilentRefresh();
     // If the user should not be forcefully logged in (e.g. if you have pages, which can be
--- a/src/app/utils/form-model.ts
+++ b/src/app/utils/form-model.ts
@@ -1,6 +1,6 @@
 import {UntypedFormControl, UntypedFormGroup, ValidatorFn} from "@angular/forms";
 
-export class FormModel<T> {
+export class FormModel<T extends {[s: string]: unknown}> {
   private readonly keys: string[];
   private readonly form: UntypedFormGroup;
   private readonly controls: {[key: string]: UntypedFormControl} = {};
@@ -32,7 +32,7 @@ export class FormModel<T> {
       this.validatorsByKey.set(key as string, []);
     }
 
-    const validators = this.validatorsByKey.get(key as string);
+    const validators = this.validatorsByKey.get(key as string)!;
     validators.push(validator);
 
     this.controls[key as string].setValidators(validators);
--- a/tsconfig.strictNullChecks.json
+++ b/tsconfig.strictNullChecks.json
@@ -11,6 +11,7 @@
     "src/app/core/**/*.ts",
     "src/app/data-structures/**/*.ts",
     "src/app/logger/**/*.ts",
+    "src/app/models/connection.model.ts",
     "src/app/sample-netzgrafik/**/*.ts",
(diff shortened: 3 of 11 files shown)


louisgreiner (commented):
I got a weird error in the console:

Image

tsconfig.strictNullChecks.json
```diff
@@ -3,6 +3,7 @@
   "compilerOptions": {
     "outDir": "./out-tsc/app",
     "types": [],
+    "strictPropertyInitialization": true,
```
emersion: I think this has been superseded by https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/pull/1298

src/app/utils/form-model.ts
```diff
@@ -1,6 +1,6 @@
 import {UntypedFormControl, UntypedFormGroup, ValidatorFn} from "@angular/forms";
 
-export class FormModel<T> {
+export class FormModel<T extends {[s: string]: any}> {
```
emersion: Can we avoid using `any` in new code? Is `unknown` enough here?

tsconfig.strictNullChecks.json
hhirtz: hm, doesn't look like it. Probably "strict" isn't inherited by the stricyNullChecks config? 🫤

src/app/utils/form-model.ts
hhirtz: Had to change some models from "interface" to "type" declarations otherwise we hit this: https://github.com/microsoft/TypeScript/issues/42825#issuecomment-780160383

tsconfig.strictNullChecks.json
hhirtz: dropped the commit

louisgreiner (commented):
Many tests are failing :/ I did not investigate why.

Rest of the PR LGTM to me!

src/app/models/filterSettings.model.ts
```diff
@@ -83,11 +83,11 @@ export class FilterSetting {
...
-      filterTrainrunCategory: null,
-      filterTrainrunFrequency: null,
-      filterTrainrunTimeCategory: null,
-      filterDirection: null,
-      filterSymmetry: null,
+      filterTrainrunCategory: [],
+      filterTrainrunFrequency: [],
+      filterTrainrunTimeCategory: [],
+      filterDirection: [],
+      filterSymmetry: [],
```
emersion: This changes behavior. `null` means "use default filter", which contains all categories/frequencies/etc. `[]` means "filter everything out".

src/app/services/auth/auth.service.ts
```diff
@@ -31,6 +31,7 @@ export class AuthService {
     location: Location,
   ) {
     if (environment.disableBackend) return;
+    if (!environment.authConfig) throw "missing authConfig from environment";
```
emersion: Can we throw `Error` objects instead of strings?

src/app/utils/form-model.ts
emersion: Interesting, didn't know about that TypeScript quirk!

src/app/data-structures/business.data.structures.ts
```diff
@@ -303,11 +303,11 @@ export interface FilterSettingDto {
...
-  filterTrainrunCategory: TrainrunCategory[]; // list of category to filter out
-  filterTrainrunFrequency: TrainrunFrequency[]; // list of frequency to filter out
-  filterTrainrunTimeCategory: TrainrunTimeCategory[]; // list of time category to filter out
-  filterDirection: Direction[]; // list of trainrun direction to filter out
-  filterSymmetry: boolean[]; // list of trainrun symmetry values (true/false) to filter out
+  filterTrainrunCategory: TrainrunCategory[] | null; // list of category to filter out
+  filterTrainrunFrequency: TrainrunFrequency[] | null; // list of frequency to filter out
+  filterTrainrunTimeCategory: TrainrunTimeCategory[] | null; // list of time category to filter out
+  filterDirection: Direction[] | null; // list of trainrun direction to filter out
+  filterSymmetry: boolean[] | null; // list of trainrun symmetry values (true/false) to filter out
```
emersion: I don't think this can happen at the DTO level: we call `initializeFilterSetting()` before `getDtos()` is called.

hhirtz: Moved the commit to another PR to discuss further... #1329

Pivouane (approved):
LGTM

louisgreiner (approved):
Thanks! LGTM

emersion (approved):

PR #1302, opened 2026-08-26
refactor: small refactor

aschantraine:
This PR is composed of small refactor, that simplify the code and clean some dead code.

### DIFF
--- a/src/app/services/data/data.service.ts
+++ b/src/app/services/data/data.service.ts
@@ -231,7 +231,7 @@ export class DataService implements OnDestroy {
       (trainrunCategory) => trainrunCategory.id === categoryId,
     );
     if (found === undefined) {
-      return this.netzgrafikDtoStore.netzgrafikDto.metadata.trainrunCategories.find(() => true);
+      return this.netzgrafikDtoStore.netzgrafikDto.metadata.trainrunCategories.at(0);
     }
     return found;
   }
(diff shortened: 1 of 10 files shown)


Pivouane (approved):
LGTM, thank you ! 🐿️

hhirtz (approved):

louisgreiner (approved):

src/app/services/data/data.service.ts
```diff
@@ -231,7 +231,7 @@ export class DataService implements OnDestroy {
       (trainrunCategory) => trainrunCategory.id === categoryId,
     );
     if (found === undefined) {
-      return this.netzgrafikDtoStore.netzgrafikDto.metadata.trainrunCategories.find(() => true);
+      return this.netzgrafikDtoStore.netzgrafikDto.metadata.trainrunCategories[0];
```
aiAdrian: What happends, when list has no entry?

aiAdrian (approved):
LGTM!

src/app/services/data/data.service.ts
aschantraine: It returns `undefined` just like the old version.

emersion: `.at(0)` is a better replacement because it returns `TrainrunCategory | undefined`, as opposed to `[0]` which returns `TrainrunCategory`.

aschantraine: Done

PR #1304, opened 2026-08-27
refactor: clean unused locals

### LINKED ISSUE #648: [Refactoring]: Quality of code: remove the unused imports/variables/methods
### Preflight Checklist

- [x] I have read the [Contributing Guidelines](https://github.com/SchweizerischeBundesbahnen/netzgrafik-editor-frontend/blob/main/CONTRIBUTING.md) for this project.
- [x] I agree to follow the [Code of Conduct](https://github.com/SchweizerischeBundesbahnen/netzgrafik-editor-frontend/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [x] I have searched the [issue tracker](https://github.com/orgs/sbb-design-systems/projects/1) for an issue that matches the one I want to file, without success.

### Request type

Request for enhancement of a component

### Functionality

Along the code, many things are unused:
- imports
- variables
- methods
- dependancies

Is there a tool that could help to clean these?

For instance:
Image

### Link to design proposal file

_No response_

aschantraine:
In this PR, I used the ts rule `noUnusedLocals` to clean the code.
Unfortunately, I couldn't manage to activate the rule since the files  generated by `openapi-generator-cli` are not respecting it.

I finally activated the typescript-eslint rule `no-unused-private-class-members` that seems to cover the same area.


louisgreiner (approved):
LGTM, many thanks for cleaning this!

I wonder if `LogService` and `LogPublishersService` could also be removed, but this is another issue.

eslint.config.js
```diff
@@ -65,6 +65,8 @@ export default defineConfig([
           varsIgnorePattern: "^_",
         },
       ],
+      "no-unused-private-class-members": "off",
```
louisgreiner, line 70:
```suggestion
      "no-unused-private-class-members": "error",
```

To enable this rule it should be `"error"`, right?

aschantraine: No, it is not intuitive but we need to disable the base rule to activate the ts-eslint rule properly...
I explained it in the commit message. Do you think we need a comment in the file ?
https://typescript-eslint.io/rules/no-unused-private-class-members/

louisgreiner: Oh crap, my bad. A comment would be nice indeed

emersion (approved):
I'm a bit surprised about the naming of the first three commits. How are these organized?

It seems like the first one almost exclusively removes unused paremeters from service constructors, except for changes in `EditorView` and `EditorMainViewComponent`. This would've been a good commit title if the commit only did that.

The two other commits don't seem to follow a pattern.

If there's no reason why these commits are split like this, I'd recommend squashing them into a single commit.

The meat of the PR looks good to me. Thanks for doing this!

PR #1306, opened 2026-08-28
refactor(services): use Map for storing trainrun sections

emersion:
There are two major kind of operations we'd like to perform on trainrun sections: get one from its ID, or iterate over all sections. A Map can fullfill both of these needs efficiently.

We currently need to convert the iterator returned by .values() to an Array, but we'll soon be able to drop this thanks to iterator helpers:
https://github.com/tc39/ecma262/pull/3395

### DIFF
--- a/src/app/services/data/trainrunsection.service.ts
+++ b/src/app/services/data/trainrunsection.service.ts
@@ -1339,7 +1339,10 @@ export class TrainrunSectionService implements OnDestroy {
     const trainrunId = trainrunMap.get(trainrunSection.trainrunId);
     const trainrun = this.trainrunService.getTrainrunFromId(trainrunId);
 
-    const newTrainrunSection: TrainrunSection = new TrainrunSection(trainrunSection);
+    const newTrainrunSection: TrainrunSection = new TrainrunSection({
+      ...trainrunSection,
+      id: undefined,
+    });
     newTrainrunSection.setTrainrun(trainrun);
 
     const sourceNodeId = nodeMap.get(trainrunSection.sourceNodeId);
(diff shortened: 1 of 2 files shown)


src/app/services/data/trainrunsection.service.ts
```diff
@@ -801,17 +803,17 @@ export class TrainrunSectionService implements OnDestroy {
...
   }
 
   getTrainrunSections(): TrainrunSection[] {
-    return Object.assign({}, this.trainrunSectionsStore).trainrunSections;
+    return [...this.trainrunSectionsStore.trainrunSections.values()];
   }
 
   getAllTrainrunSectionsForTrainrun(trainrunID: number): TrainrunSection[] {
-    const trainrunSections = this.trainrunSectionsStore.trainrunSections.filter(
+    const trainrunSections = [...this.trainrunSectionsStore.trainrunSections.values()].filter(
```
aiAdrian: This change improves lookups by section id, but not lookups by trainrunId. getAllTrainrunSectionsForTrainrun() still has to iterate over all sections and filter them, so the complexity remains O(n).

The new Map does not help for this access pattern since trainrunId is not the map key. The original performance issue is querying all sections belonging to a specific train run, and this method still performs a full scan for that use case.

aiAdrian: Image

aiAdrian: @emersion -> the PR goes into the correct direction

emersion:
Yes, you're correct that this PR and https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/pull/1248 address different performance bottlenecks, and that ideally we'd fix both.

PR #1307, opened 2026-08-28
refactor:chore: eliminate explicit any

### LINKED ISSUE #1126: Eliminate `any` types

aschantraine:
This PR clean the last occurrences of `any` in the codebase and enable the ts-eslint rule `no-explicit-any`.

### DIFF
--- a/src/app/view/editor-tools-view-component/editor-tools-view.component.ts
+++ b/src/app/view/editor-tools-view-component/editor-tools-view.component.ts
@@ -80,7 +80,7 @@ export class EditorToolsViewComponent {
     const file = fileInput.files[0];
     const reader = new FileReader();
     reader.onload = () => {
-      let netzgrafikDto: any;
+      let netzgrafikDto: NetzgrafikDto;
       try {
         netzgrafikDto = JSON.parse(reader.result.toString());
       } catch {
@@ -240,7 +240,10 @@ export class EditorToolsViewComponent {
     });
     const url = window.URL.createObjectURL(blob);
 
-    const nav = window.navigator as any;
+    const nav = window.navigator as Navigator & {
+      msSaveOrOpenBlob?: unknown;
+      msSaveBlob?: (blob: Blob, filename: string) => void;
+    };
     if (nav.msSaveOrOpenBlob) {
       nav.msSaveBlob(blob, filename);
     } else {
--- a/src/app/view/variant/variants-view/variants-view.component.spec.ts
+++ b/src/app/view/variant/variants-view/variants-view.component.spec.ts
@@ -10,7 +10,7 @@ import {
   VariantControllerBackendService,
   VersionControllerBackendService,
 } from "../../../api/generated";
-import {of} from "rxjs";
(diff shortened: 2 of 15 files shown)


src/app/view/variant/variants-view/variants-view.component.spec.ts
```diff
@@ -22,8 +22,8 @@ describe("VariantsViewComponent", () => {
   let component: VariantsViewComponent;
   let fixture: ComponentFixture<VariantsViewComponent>;
 
-  let projectControllerBackendService: Partial<ProjectControllerBackendService>;
-  let variantControllerBackendService: Partial<VariantControllerBackendService>;
+  let projectControllerBackendService: {getProject: () => Observable<ProjectDto>};
+  let variantControllerBackendService: {createVariant: () => Observable<number>};
```
emersion, line 26: We loose the synchronization with the backend service type, but I don't see a better way while still keeping things simple (due to overloads in the method definition).

src/app/view/editor-tools-view-component/editor-tools-view.component.ts
```diff
@@ -240,7 +240,10 @@ export class EditorToolsViewComponent {
     });
     const url = window.URL.createObjectURL(blob);
 
-    const nav = window.navigator as any;
+    const nav = window.navigator as Navigator & {
+      msSaveOrOpenBlob?: unknown;
+      msSaveBlob?: (blob: Blob, filename: string) => void;
+    };
```
emersion, line 246: Ideally we should just get rid of this Internet Explorer specific API, but for now this is fine.

emersion (approved):
Very nice, looks good!

Feel free to resolve comments below, these are just notes.

src/app/view/editor-tools-view-component/editor-tools-view.component.ts
hhirtz: Also testing `nav.msSaveOrOpenBlob` and then using `nav.msSaveBlob` triggers a null warning 🤷 (whatever)

src/benchmarks/port-ordering.benchmark.spec.ts
```diff
@@ -159,7 +160,7 @@ const loadAlphabetical = (
   nodeService: NodeService,
   dto: unknown,
 ): Node[] => {
-  dataService.loadNetzgrafikDto(structuredClone(dto) as any);
+  dataService.loadNetzgrafikDto(structuredClone(dto) as NetzgrafikDto);
```
hhirtz:
```suggestion
  dto: NetzgrafikDto,
): Node[] => {
  dataService.loadNetzgrafikDto(structuredClone(dto));
```

hhirtz (changes requested):

src/benchmarks/port-ordering.benchmark.spec.ts
aschantraine: Done

hhirtz (approved):

PR #1313, opened 2026-08-28
chore(deps): upgrade ESLint to v10 

emersion:
_See individual commits._

Of all the breaking changes, only the recommended rule set update seems to affect us: https://eslint.org/docs/latest/use/migrate-to-10.0.0#breaking-changes-for-users


hhirtz (approved):
the strictNullChecks config also seems to pass

SharglutDev (approved):
Surprisingly smooth migration, thanks !

louisgreiner (approved):
Thanks again! ✅

PR #1322, opened 2026-09-01
Move multiple methods from TrainrunSectionViewObject

louisgreiner:
_See individual commits_

### DIFF
--- a/src/app/services/data/trainrunsection.service.ts
+++ b/src/app/services/data/trainrunsection.service.ts
@@ -1644,12 +1644,15 @@ export class TrainrunSectionService implements OnDestroy {
     return numberOfCollapsedStops;
   }
 
-  getTrainrunSectionsGroupOrientedBasedOnPort(port: Port): TrainrunSection[] | undefined {
-    const section = port.getTrainrunSection();
+  getTrainrunSectionGroupForSection(section: TrainrunSection): TrainrunSection[] {
     const sections = this.getAllTrainrunSectionsForTrainrun(section.getTrainrun().getId());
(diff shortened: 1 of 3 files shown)


src/app/services/data/trainrunsection.service.ts
```diff
@@ -1644,11 +1644,15 @@ export class TrainrunSectionService implements OnDestroy {
     return numberOfCollapsedStops;
   }
 
-  getTrainrunSectionsGroupOrientedBasedOnPort(port: Port): TrainrunSection[] | undefined {
-    const section = port.getTrainrunSection();
+  getTrainrunSectionGroupForSection(section: TrainrunSection): TrainrunSection[] {
     const sections = this.getAllTrainrunSectionsForTrainrun(section.getTrainrun().getId());
     const groups = this.groupTrainrunSectionsIntoChains(sections);
     const group = groups.find((group) => group.some((trs) => trs.getId() === section.getId()));
+    return group ?? [section];
```
hhirtz, line 1651: This null fallback changes the behavior of `getTrainrunSectionsGroupOrientedBasedOnPort`, which cannot return undefined anymore.

is this on purpose?

louisgreiner: You're right, that was an oversight, nice spot. Fixed it.

hhirtz: the `group === undefined` check below can be removed as well

louisgreiner: True, removed it also.

hhirtz (approved):

Math-R (approved):
lgtm

PR #1323, opened 2026-09-01
Add `allowFirstAndLastToBeNonStop` option to `groupTrainrunSectionsIntoChains()`

louisgreiner:
The option is not used yet, but will soon be:
- if `allowFirstAndLastToBeNonStop === true`: the first and last
  sections of the group can respectively stop or not at the first
  and last non-collapsed node.
- if `allowFirstAndLastToBeNonStop === false`: the first and last
  sections of the group necesseraly stop respectively at the first
  and last non-collapsed node.

Goes after https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/pull/1322

### DIFF
--- a/src/app/services/data/trainrunsection.service.ts
+++ b/src/app/services/data/trainrunsection.service.ts
@@ -1558,10 +1558,15 @@ export class TrainrunSectionService implements OnDestroy {
    * Groups consecutive TrainrunSections that have collapsed nodes between them
    * into chains. Each chain starts and ends with a non-collapsed node.
    * Start and end nodes can be accessed via: sections[0].getSourceNode() and sections[sections.length - 1].getTargetNode()
-   * @param trainrunSections List of TrainrunSections to group
-   * @returns Array of section chains
+   * @param trainrunSections List of TrainrunSections
+   * @param includeExpandedNonStopNodes if true, then section groups include
+   * non-stop nodes, even when those nodes aren't collapsed.
+   * @returns Array of sections group
    */
-  groupTrainrunSectionsIntoChains(trainrunSections: TrainrunSection[]): TrainrunSection[][] {
+  groupTrainrunSectionsIntoChains(
+    trainrunSections: TrainrunSection[],
+    includeExpandedNonStopNodes: boolean,
+  ): TrainrunSection[][] {
     const groups: TrainrunSection[][] = [];
     const visitedSections = new Set<number>();
 
@@ -1574,7 +1579,12 @@ export class TrainrunSectionService implements OnDestroy {
         section.getTargetNode(),
         section,
       );
-      while (backwardIterator.hasNext() && backwardIterator.current().node.getIsCollapsed()) {
+      while (
+        backwardIterator.hasNext() &&
+        (backwardIterator.current().node.getIsCollapsed() ||
+          (includeExpandedNonStopNodes &&
+            backwardIterator.current().node.isNonStop(backwardIterator.current().trainrunSection)))
+      ) {
         backwardIterator.next();
       }
       const startNode = backwardIterator.current().node;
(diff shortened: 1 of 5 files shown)


src/app/services/data/trainrunsection.service.ts
```diff
@@ -1558,10 +1558,20 @@ export class TrainrunSectionService implements OnDestroy {
    * Groups consecutive TrainrunSections that have collapsed nodes between them
    * into chains. Each chain starts and ends with a non-collapsed node.
    * Start and end nodes can be accessed via: sections[0].getSourceNode() and sections[sections.length - 1].getTargetNode()
+   * @param allowFirstAndLastToBeNonStop Depending on:
+   * - if `allowFirstAndLastToBeNonStop === true`: the first and last
+   *   sections of the group can respectively stop or not at the first
+   *   and last non-collapsed node.
+   * - if `allowFirstAndLastToBeNonStop === false`: the first and last
+   *   sections of the group necesseraly stop respectively at the first
+   *   and last non-collapsed node.
```
hhirtz: hmm i might not understand well but if allowFirstAndLastToBeNonStop is true then section chains/groups may also contain non-stop, non-collapsed nodes anywhere, not just first and last (?)

how about

```suggestion
   * @param dontSplitOnNonStopNodes if true, then section groups are not
   * split between non-stop nodes, even when those nodes aren't collapsed.
```

Ideally, instead of adding booleans to tweak when to split, this function would take a callback function `(pair: TrainrunSectionNodePair) => bool` that returns true when the input section array must be split on the given pair's node.

louisgreiner: > non-collapsed nodes anywhere

I don't think that's true.

The idea is that:
- (default case since now) if `allowFirstAndLastToBeNonStop === false`, then groups are:
  - `[expanded stop node] ... [collapsed {stop/non-stop} node] ... [expanded stop node]`
- if `allowFirstAndLastToBeNonStop === true`, then groups are:
  - `[expanded {stop/non-stop} node] ... [collapsed {stop/non-stop} node] ... [expanded {stop/non-stop} node]`

But I'm ok for renaming it, I could not manage to think to something relevant, what would you propose?

louisgreiner: I did not understand well the callback function thing

louisgreiner: Renamed it in `includeExpandedNonStopNodes`.

louisgreiner: I'm not sure using a callback function is needed here, as we have only 2 types of group. I would leave it like that for now if you're ok with that.

hhirtz: yeah for now it's fine. `includeExpandedNonStopNodes` is a good name. I still think the doc comment is too convoluted and can be replaced with the suggestion above

louisgreiner: Right, changed it

hhirtz (approved):
thank you, the rest of the PR looks good to me

src/app/services/data/trainrunsection.service.ts
```diff
@@ -1644,19 +1657,40 @@ export class TrainrunSectionService implements OnDestroy {
...
+      includeExpandedNonStopNodes,
+    );
     if (group[0].getSourcePortId() === port.getId()) {
       return group;
     } else {
       return [...group].reverse();
     }
   }
+
+  getDirectedTrainrunSectionProxiesGroup(
```
Math-R: Does this function should be in https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/pull/1324 ?

Math-R (commented):
I'm a bit concerned about the drilling of the `includeExpandedNonStopNodes` parameter.

It looks like it's being passed down through quite a lot of functions.

Also, maybe I missed it, but I couldn't find anywhere where this parameter is set to `true`, either in this PR or in [[this one](https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/pull/1324/changes)]

src/app/services/data/trainrunsection.service.ts
louisgreiner: Nice spot, it's bad split between this PR and the next (https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/pull/1324). I've removed the change in this PR.

Math-R (approved):
lgtm

Issue #1285, opened 2026-08-25
Migrate to Vitest away from Karma


emersion:
Angular v22 has an optional migration:

> Migrate projects using legacy Karma unit-test builder to the new unit-test builder with Vitest.

We should apply it in a dedicated PR.

See https://angular.dev/guide/testing/migrating-to-vitest

PR #1327, opened 2026-09-01
chore: explicitly import jasmine types for test files

louisgreiner:
The jasmine types (`describe/it/expect`) were not provided by the former config, resulting in error in IDEs.

This solves the missing types from jasmin in my IDE. I don't know if there is an alternative.

Before:
image

After:
image

### DIFF
--- a/src/test.ts
+++ b/src/test.ts
@@ -3,6 +3,7 @@
 import "zone.js/testing";
 import {getTestBed} from "@angular/core/testing";
 import {BrowserTestingModule, platformBrowserTesting} from "@angular/platform-browser/testing";
+import type {} from "jasmine";
 
 // First, initialize the Angular testing environment.
 getTestBed().initTestEnvironment(BrowserTestingModule, platformBrowserTesting(), {


tsconfig.json
```diff
@@ -13,6 +13,7 @@
     "importHelpers": true,
     "target": "ES2022",
     "typeRoots": ["node_modules/@types"],
+    "types": ["jasmine", "node"],
```
hhirtz: Adding this line excludes all other `@types` packages in node_modules.

It seems the workaround is to add `import "jasmine"` to all spec files but... yeah.. 
https://github.com/angular/angular-cli/issues/8909

emersion (commented):
This makes all jasmine exports available in non-test files. `tsconfig.spec.json` already contains these lines and targets only test files.

I don't think relying on implicit globals is great anyways, so I wouldn't mind explicitly pulling these from an import.

Somewhat related: https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/issues/1285

hhirtz (approved):

src/app/services/util/trainrunsection.helper.spec.ts
```diff
@@ -16,6 +16,7 @@ import {NetzgrafikUnitTesting} from "../../../integration-testing/netzgrafik.uni
 import {FilterService} from "../ui/filter.service";
 import {NetzgrafikColoringService} from "../data/netzgrafikColoring.service";
 import {TrainrunsectionHelper} from "./trainrunsection.helper";
+import {} from "jasmine";
```
emersion: Is this something we could add to `src/test.ts` instead?

Does `import "jasime"` work too?

emersion (commented):
Can the PR title and commit message be updated to reflect the new approach?

src/app/services/util/trainrunsection.helper.spec.ts
louisgreiner: yep it does! that's way better indeed

SharglutDev (approved):
Lgtm 👍

emersion (approved):

PR #1333, opened 2026-09-03
ci: deploy github pages only in main

louisgreiner:
This disables the job when merging PR into other branches than main, for instance for feature branches.

This should fix this for instance: https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/actions/runs/33763890042/job/100677510824

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -122,7 +122,7 @@ jobs:
           push: false
 
   deploy-github-pages:
-    if: github.event_name == 'push'
+    if: github.event_name == 'push' && github.ref == 'refs/heads/main'
     needs: build-standalonedemo
     permissions:
       pages: write


Math-R (approved):

.github/workflows/build.yml
emersion, line 5: This line already only runs the workflow on `main` branch pushes only.

louisgreiner: Oh, the issue was because of this actually: https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/commit/f58fda29144f564b0862e891e5dc26c00f0dd41b

I pushed (by mistake, if this doesn't fit to you i'll remove it) this instead: https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/commit/09146ee571b69f920d6c6f90cd08e14e3d520f9a

If that's ok, i'm closing this PR.

emersion: Yup, that fixup looks good to me!

Issue #1289, opened 2026-08-25
Switch TypeScript strict mode on


emersion:
Right now, `strict` is off in our TypeScript configuration. Enable it. See https://www.typescriptlang.org/tsconfig/#strict

- [ ] Fix all `strictNullChecks` errors
- [ ] Fix all `strictFunctionTypes` errors

This requires migration the whole app to `strictNullChecks` (see initial work on this in https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/issues/645).

PR #1341, opened 2026-09-07
refactor: enable TypeScript's `strictFunctionTypes` option

emersion:
This option enables more checks around function parameters: https://www.typescriptlang.org/tsconfig/#strictFunctionTypes

d3 is the main error culprit. Some types need to be narrowed down (e.g. switched from `SVGElement` to `SVGGElement`, notice the extra "G"). `selectAll()` can't guess what the type of the element and datum will be, so we need to explicitly specify these.

Unfortunately, `d3.Selection` cannot be easily converted from a more precise type to a more general type. This requires us to sometimes use a more general type in `append()`, and for one call we need a dummy `filter()` call to switch the element type.

References: https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/issues/1289

### DIFF
--- a/src/app/view/editor-main-view/data-views/nodes.view.ts
+++ b/src/app/view/editor-main-view/data-views/nodes.view.ts
@@ -27,7 +27,7 @@ type NodeDragEvent = d3.D3DragEvent<SVGElement, NodeViewObject, unknown>;
 
 export class NodesView {
   dragPreviousMousePosition: Vec2D;
-  nodeGroup: d3.Selection<SVGElement, undefined, Element, undefined>;
+  nodeGroup: d3.Selection<SVGGElement, undefined, Element, undefined>;
   draggable: d3.DragBehavior<SVGElement, NodeViewObject, unknown>;
   dragDomObj: SVGElement | null = null;
 
@@ -40,7 +40,7 @@ export class NodesView {
     this.dragPreviousMousePosition = new Vec2D();
   }
 
-  setGroup(nodeGroup: d3.Selection<SVGElement, undefined, Element, undefined>) {
+  setGroup(nodeGroup: d3.Selection<SVGGElement, undefined, Element, undefined>) {
     this.nodeGroup = nodeGroup;
     this.nodeGroup.attr("class", "NodesView");
   }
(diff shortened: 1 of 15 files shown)


src/app/view/editor-main-view/data-views/nodes.view.ts
```diff
@@ -626,7 +630,10 @@ export class NodesView {
       return;
     }
 
+    // Note, an awkward filter() call is required here to generalize the
+    // selection type and make TypeScript happy
     added
+      .filter<SVGElement>(() => true)
```
hhirtz, line 636: It's not required, we can cast the type and avoid the runtime cost:


```suggestion
    (added as unknown as d3.Selection<SVGElement, NodeViewObject, Element, undefined>)
```

It's not different than the other unchecked type annotations in this file.

hhirtz (commented):
d3.Selection is invariant but at the same time allows users to just cast between element types.. why the generic arg at this point?

aiAdrian (approved):
LGTM! Thank you for this valuable work(change).

SarahBellaha (approved):

src/app/view/editor-main-view/data-views/nodes.view.ts
emersion: I'd prefer to avoid the cast if possible. This ensures only the element type is overwritten, leaving the rest of the generic arguments untouched (e.g. datum).

(The runtime cost is negligible.)

hhirtz: yeah it's a good reason to use "filter", but the comment above is still wrong. if you prefer filter then:

```suggestion
    // Note, an awkward filter() call is used here to generalize the
    // selection type and make TypeScript happy
    added
      .filter<SVGElement>(() => true)
```

or

```suggestion
    // Note, an awkward filter() call is used here to generalize the
    // selection type and make TypeScript happy. Contrary to a cast, this ensures only the element type is overwritten, leaving the rest of the generic arguments untouched 
    added
      .filter<SVGElement>(() => true)
```

but i guess feel free to just resolve and merge...

hhirtz (approved):

src/app/view/editor-main-view/data-views/nodes.view.ts
emersion: Something is required to adjust the type, be it a cast or a `.filter()`, so I don't think the comment is wrong.

Issue #999, opened 2026-05-11
Migrate to standalone components


emersion:
We're using `standalone: false` in quite a few spots. This is a bad practice.

See https://github.com/angular-eslint/angular-eslint/blob/main/packages/eslint-plugin/docs/rules/prefer-standalone.md

aiAdrian:
Nice spot

PR #1343, opened 2026-09-08
chore: migrate to Vitest

### LINKED ISSUE #1285: Migrate to Vitest away from Karma

emersion:
Migration guide: https://angular.dev/guide/testing/migrating-to-vitest

_See individual commits._

Because our components are not standalone (see #999), we need to include `src/main.ts` in `tsconfig.spec.json` so that TypeScript is aware of all declared components: https://github.com/angular/angular/issues/65665#issuecomment-3582136579

We also need to disable AOT compilation for tests: AOT requires standalone components: https://github.com/angular/angular/issues/65665#issuecomment-3853223988

We're migrating from a browser environment to a Node environment with jsdom providing emulation for Web APIs. jsdom is missing an implementation for window.matchMedia, so we need to emulate it manually. See: https://jestjs.io/docs/manual-mocks#mocking-methods-which-are-not-implemented-in-jsdom

Node support for localStorage may interfere with tests. `NODE_OPTIONS=--no-webstorage` can be set to work around this issue. See: https://github.com/vitest-dev/vitest/issues/8757

Switch from jasmine globals to vitest in `tsconfig.spec.json`. This requires adding "node_modules" to "typeRoots". See: https://vitest.dev/config/globals.html#globals

Vitest is stricter than Jasmine when checking `.toContain()` for equality. We need to switch to the more lax `.toContainEqual()` variant.

Vitest doesn't have `.withContext()`. Instead, provide a description string as `expect()`'s second argument.

Closes: https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/issues/1285

Follow-up task: migrate tested components to standalone, and enable AOT compilation for tests (WIP PR: https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/pull/1357).


package.json
```diff
@@ -28,16 +28,14 @@
     "build": "ng run netzgrafik-frontend:ngsscbuild",
     "build:standalone": "ng build --configuration=standalone && tsc --project tsconfig.defs.json",
     "build:standalonedemo": "ng build --configuration=standalonedemo",
-    "test": "ng test -c ci",
+    "test": "ng test",
```
louisgreiner, line 31: Launching the tests outputs this error multiple times:

```
stderr | src/app/view/project/project-dialog/project-form/project-form.component.spec.ts > ProjectFormComponent > should create
Could not find @sbb-esta/angular typography. Most @sbb-esta/angular components may not work as expected.

stderr | src/app/view/project/project-dialog/project-form/project-form.component.spec.ts > ProjectFormComponent > validEMailExamples test
Could not find @sbb-esta/angular typography. Most @sbb-esta/angular components may not work as expected.

stderr | src/app/view/project/project-dialog/project-form/project-form.component.spec.ts > ProjectFormComponent > invalidEMailExamples test
Could not find @sbb-esta/angular typography. Most @sbb-esta/angular components may not work as expected.
```

I guess this was already to case before this PR?

package.json
```diff
@@ -28,16 +28,14 @@
...
     "format": "oxfmt",
     "format:check": "oxfmt --check",
     "lint": "ng lint",
     "lint:fix": "ng lint --fix",
     "lint:null": "tsc --project tsconfig.strictNullChecks.json",
     "generate:api": "openapi-generator-cli generate",
-    "local-test": "ng test --watch --code-coverage",
-    "extract-i18n": "ng extract-i18n",
-    "benchmark": "ng test -c benchmark"
+    "extract-i18n": "ng extract-i18n"
```
louisgreiner, line 38: The port-ordering benchmark was formerly not part of the tests, but this change re-introduces it. I'm ok with that, since the tests are very fast.

Was it intended from you?

package.json
emersion, line 31: The README should be updated to reflect the new command syntax. For instance, `--browsers` is no more.

package.json
emersion: Ah, definitely not intended. I'll look how to keep benchmarks separate with Vitest!

package.json
emersion: Good question, maybe this is due to the browser-less test environment, will investigate.

PR #1345, opened 2026-09-08
fix: avoid ID collision in `createNewTrainrunSectionFromDto()`

emersion:
This function invokes the TrainrunSection constructor with a fixed ID. If the trainrun section service already has another trainrun section with the same ID, we end up with two different sections sharing the same ID.

Avoid this by making the `id` field optional in the constructor, and erasing it in `createNewTrainrunSectionFromDto()`.

### DIFF
--- a/src/app/models/trainrunsection.model.ts
+++ b/src/app/models/trainrunsection.model.ts
@@ -26,6 +26,9 @@ const EMPTY_TEXT_POSITIONS: TrainrunSectionTextPositions = {
   [TrainrunSectionText.TrainrunSectionNumberOfStops]: {x: 0, y: 0},
 };
 
+type TrainrunSectionOptions = Omit<TrainrunSectionDto, "id"> &
+  Partial<Pick<TrainrunSectionDto, "id">>;
+
 export class TrainrunSection {
   private static currentId = 0;
 
@@ -61,7 +64,7 @@ export class TrainrunSection {
 
   constructor(
     {
-      id,
+      id = TrainrunSection.incrementId(),
       sourceNodeId,
       sourcePortId,
       targetNodeId,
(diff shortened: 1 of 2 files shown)


aiAdrian (approved):
LGTM!

src/app/models/trainrunsection.model.ts
```diff
@@ -26,6 +26,9 @@ const EMPTY_TEXT_POSITIONS: TrainrunSectionTextPositions = {
   [TrainrunSectionText.TrainrunSectionNumberOfStops]: {x: 0, y: 0},
 };
 
+type TrainrunSectionOptions = Omit<TrainrunSectionDto, "id"> &
```
DucNg, line 29: Where is this type used?

src/app/models/trainrunsection.model.ts
```diff
@@ -83,8 +86,7 @@ export class TrainrunSection {
         textPositions: {...EMPTY_TEXT_POSITIONS},
       },
       warnings,
-    }: TrainrunSectionDto = {
-      id: TrainrunSection.incrementId(),
```
DucNg, line 87: Why isn't this already ensured each id were unique?

DucNg (approved):
LGTM

louisgreiner (approved):
Nice spot. Have you encountered any bug related to that? Or it's only to prevent?

Because as this method is called when loading a DTO, before that the objects are flushed I think, making this case not happening. But I agree this is better to prevent anyway.

src/app/models/trainrunsection.model.ts
emersion: In the constructor right below.

src/app/models/trainrunsection.model.ts
emersion: We invoke the constructor in two cases:

1. When loading a JSON file. In that case, we want to retain all IDs from the JSON file.
2. When pasting copied nodes. In that case, we want to generate new IDs.

This class is a model and doesn't have access to the rest of the nodes: it can't check for uniqueness.

SarahBellaha (approved):

PR #1348, opened 2026-09-09
chore: release 2.12.0-rc1

louisgreiner:
Release-As: 2.12.0-rc1

According to https://pkg.go.dev/github.com/googleapis/release-please#readme-how-do-i-change-the-version-number, this should trigger a new version for the release candidate.

This will (from my understanding) replace https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/pull/1030


emersion (approved):
I'm not 100% sure this will do what we want, but it seems like it should according to the docs. Let's try!

PR #1349, opened 2026-09-09
fix(view): require mousedown event inside view to start panning

### LINKED ISSUE #1347: Network graph moves even if the user has began to click outside of the viewport
### What is the input mode?

Mouse

### What type of issue is it?

Functionality

### What happened?

When the user clicks outside of the application, keeps the click down and goes back to the viewport, the network graph moves. It shoudn't. We should ensure that the user's click is in the viewport, if not we don't move.

https://github.com/user-attachments/assets/12d89101-870b-485c-b520-246fb3ac53ce

Check that mouse down has happened in the viewport before allowing to move the network graph

video

Local

Firefox

2.11.1

emersion:
If the user presses the mouse button on the header bar or outside of the app (e.g. on the browser's UI), don't pan.

This prevents surprising behavior when interacting with other UI elements and moving the mouse over NGE. NGE should not pan if an [implicit grab][1] has been started from the outside.

[1]: https://wayland.freedesktop.org/docs/book/Protocol.html#input

Closes: https://github.com/OpenRailAssociation/netzgrafik-editor-frontend/issues/1347

### DIFF
--- a/src/app/view/util/svg.mouse.controller.ts
+++ b/src/app/view/util/svg.mouse.controller.ts
@@ -28,6 +28,11 @@ export class SVGMouseController {
   private onPanning = false;
   private viewboxIsFixed = false;
 
+  /**
+   * True if a mouse button has been pressed over the drawing context, and is
+   * still being pressed.
+   */
+  private buttonDown = false;
   private lastMouseEventTimeStamp: number = undefined;
   private ctrlKeyPressed = false;
 
@@ -221,6 +226,8 @@ export class SVGMouseController {
   }
 
   private onGraphContainerMousedown(event: MouseEvent) {
+    this.buttonDown = true;
+
     // reset to initial value (start panning)
     if (event.shiftKey || (event.buttons === 2 && this.lastMouseEventTimeStamp === undefined)) {
       this.previousMultiSelectShiftPosition = this.getCurrentMousePosition(event);
@@ -247,7 +254,7 @@ export class SVGMouseController {
       if (this.previousMultiSelectShiftPosition === null) {
         if (this.onPanning) {
           this.panDrawingContext(event);
-        } else if (event.button === 0) {
+        } else if (event.button === 0 && this.buttonDown) {
           if (this.previousPanMousePosition !== null) {
             const delta = Vec2D.norm(
               Vec2D.sub(this.previousPanMousePosition, this.getCurrentMousePosition(event)),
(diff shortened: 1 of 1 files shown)


louisgreiner (approved):
Working perfectly, thanks!

achrafmohye (approved):

src/app/view/util/svg.mouse.controller.ts
```diff
@@ -28,6 +28,7 @@ export class SVGMouseController {
   private onPanning = false;
   private viewboxIsFixed = false;
 
+  private buttonDown = false;
```
hhirtz, line 35: it would be nice if more variables had comments. i have a hard time reading through the code base 

```suggestion
  /** True between a mousedown and a mouseup event */
  private sawMouseDown = false;
```

the name change is a nit

src/app/view/util/svg.mouse.controller.ts
```diff
@@ -247,7 +250,7 @@ export class SVGMouseController {
       if (this.previousMultiSelectShiftPosition === null) {
         if (this.onPanning) {
           this.panDrawingContext(event);
-        } else if (event.button === 0) {
+        } else if (event.button === 0 && this.buttonDown) {
```
hhirtz, line 257: shouldn't the buttonDown test be on the first if (buttons > 0) ?

hhirtz (approved):

src/app/view/util/svg.mouse.controller.ts
Math-R: [event.button can take 0, 1 , 2](https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/button#value). It is relative to which button triggered the event.

Math-R (approved):
lgtm

src/app/view/util/svg.mouse.controller.ts
emersion: That would require more changes, because the `if (buttons > 0)` has an `else` branch that would need a refactor if we added a new condition.

Behavior is identical regardless, since the rest is guarded by `if (previousMultiSelectShiftPosition)`, which is only populated by `onGraphContainerMousedown()`.

src/app/view/util/svg.mouse.controller.ts
emersion: I've added a doc comment.

